PR #24, opened 2015-02-13
Bump to 3.12.1

tkelman:
run `./get.ASL` so command-line ipopt ampl executable will be available
remove `--enable-dependency-linking` since it's on by default now
run make test, and install with `-j1` (people might set `ENV["MAKEFLAGS"]`)
add `-lgfortran` to `--with-blas` (needed for linking the ipopt ampl executable, apparently)

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -5,7 +5,7 @@ using BinDeps
 windllname = "IpOpt-vc10"
 libipopt = library_dependency("libipopt", aliases=[windllname])
 
-ipoptname = "Ipopt-3.11.7"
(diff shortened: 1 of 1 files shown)


mlubin:
Thanks!

tkelman:
There's a chance someone might want to use a non-gfortran compiler at some point. If that comes up we can probably revisit the static linking of BLAS that we're doing here, and set a Julia 0.3.3 minimum version requirement.

mlubin:
Yeah, that doesn't seem like a big issue for now.

Issue #27, opened 2015-02-25
Can a newer version of Ipopt be distributed for windows?


davidanthoff:
I currently have a JuMP problem that solves on juliabox, but not on my local (windows) machine. The only difference between the two setups that I was able to find is the version of Ipopt: juliabox is on 3.11.7, whereas my local machine is on 3.11.0 (no custom install, just what Ipop.jl brings in the box).

Could the bundled Ipopt version for windows be updated?

Also, if someone from the Ipopt.jl or JuMP team wants to try this out, I'd be happy to give you access to the bitbucket repo with the replication code. Thanks!

@tkelman, @mlubin

davidanthoff:
Oh, and I should say, the problem on my windows machine is that I get a maximum number of iterations exceeded.

tkelman:
Yeah I have the latest release up on my personal project on the opensuse build service that you can download with WinRPM via
```
using WinRPM
push!(WinRPM.sources, "http://download.opensuse.org/repositories/home:/kelman:/mingw-coinor/openSUSE_13.1")
WinRPM.update()
WinRPM.install("Ipopt")
```
I've been waiting for the next release of https://github.com/ampl/mp so we can include the command-line AMPL executable as well before submitting to the official default repository.

davidanthoff:
I executed all of these commands but when I run my problem it is still using Ipopt 3.11.0. Is there anything else I need to do? Your WinRPM commands actually don't replace the Ipopt binaries in the Ipopt.jl package directory, or do they?

tkelman:
No, you'd need to adjust the `deps/build.jl` file and rerun `Pkg.build("Ipopt")`, or change the library path in `deps/deps.jl` (and *don't* rerun `Pkg.build("Ipopt")`). See the tk/winrpm branch for the former.

davidanthoff:
I checked out the tk/winrpm branch for the Ipopt package, and then did a Pkg.build("Ipopt"). I get the following output/error:
````
julia> Pkg.build("Ipopt")
INFO: Building WinRPM
INFO: Downloading http://download.opensuse.org/repositories/windows:/mingw:/win32/openSUSE_13.1//repodata/repomd.xml
INFO: Downloading http://download.opensuse.org/repositories/windows:/mingw:/win64/openSUSE_13.1//repodata/repomd.xml
INFO: Building Ipopt
[...]

PR #29, opened 2015-02-25
RFC: Switch to a WinRPM binary

### LINKED ISSUE #7: 32/64-bit switch isn't seamless on windows
When switching between 32- and 64-bit julia executables on Windows, ``Pkg.build("Ipopt")`` isn't sufficient to get the proper library set up. I had to manually clear ``deps/usr/lib`` and then run ``Pkg.build("Ipopt")``.

tkelman:
add WinRPM to REQUIRE on Windows

Would fix #7 and #27. Should wait until it's submitted to the official repository instead of my personal project before merging. I'm waiting for the next ampl-mp release before submitting that (for the command-line executable), but reference blas and lapack, and mumps 4.10 are already there.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,3 +3,4 @@ BinDeps
 MathProgBase 0.3.7 0.4.0-
 @osx Homebrew
 Compat
+@windows WinRPM
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -66,9 +66,6 @@ end
 # Windows
 @windows_only begin
     using WinRPM
-    # TODO: remove the next 2 lines once package is submitted to official repository
-    push!(WinRPM.sources, "http://download.opensuse.org/repositories/home:/kelman:/mingw-coinor/openSUSE_13.1")
-    WinRPM.update()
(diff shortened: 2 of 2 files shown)


davidanthoff:
@tkelman Is this ready to be merged as the default, or are you still waiting for the ampl-mp release?

tkelman:
Thanks for the bump. I'll submit ampl-mp and ipopt to the official repo now, see how long that takes.

davidanthoff:
Great, thanks!

tkelman:
@mlubin @IainNZ thoughts? We should probably benchmark but I don't have much time to do that right now. "Able to solve at all" is certainly a step up from "hits iteration limit," in the case of @davidanthoff's particular problem.

mlubin:
I'm okay with merging this. Better to have a default that works reasonably well even if it's not well optimized.

tkelman:
For lack of any objections, will merge soon then?

davidanthoff:
Great, thanks! Could you also tag a new version?

tkelman:
https://github.com/JuliaLang/METADATA.jl/pull/2415

PR #35, opened 2015-05-17
save the location of the ampl executable

tkelman:
for the sake of AmplNLWriter.jl, cc @JackDunnNZ

Any alternate suggestions for what to name this variable? Could just use `Ipopt.ipopt` maybe?

### DIFF
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -5,6 +5,7 @@ if isfile(joinpath(dirname(@__FILE__),"..","deps","deps.jl"))
 else
     error("Ipopt not properly installed. Please run Pkg.build(\"Ipopt\")")
 end
+amplexe = joinpath(dirname(libipopt), "..", "bin", "ipopt")
 
 using Compat
 


JackDunnNZ:
Hmm, `Ipopt.ipopt` is nice for consistency with CoinOptServices.jl, but doesn't really make sense in isolation.

We have `AmplNLSolver(CoinOptServices.bonmin)`, so maybe `AmplNLSolver(Ipopt.ipopt)` is nicer than `AmplNLSolver(Ipopt.amplexe)` or simllar? Not really sure

tkelman:
opinions, @JuliaOpt/nonlinear-solvers @joehuchette @IainNZ?

mlubin:
I wouldn't worry about the naming here too much. AmplNLWriter.jl could export its own names like ``IpoptNLSolver()`` if the relevant packages are installed.

tkelman:
That's a good idea. Should we just merge this then and let AmplNLWriter worry about presenting more user-friendly and consistent names? I have `OsilBonminSolver` and `OsilCouenneSolver` in CoinOptServices, but you don't have to follow my lead of which order to put file format vs solver in the name.

JackDunnNZ:
Yep that sounds good

PR #37, opened 2015-05-28
add restoration phase options for MPB

odow:
Allows the user to set ``resto.[option]`` options. 

Not sure if there is a better/preferred way of handling things like this?


mlubin:
LGTM

IainNZ:
I'll tag a new version

IainNZ:
Still awake @odow ?!

odow:
I'm in Denmark at DTU until the end of July so its actually 5pm here right now

IainNZ:
Oh ok, that makes more sense!

PR #45, opened 2015-10-20
Configure Ipopt with -DNO_fpu_control on ARM

jgonzal13:
ref https://projects.coin-or.org/BuildTools/wiki/current-issues (says "On P-Series(ppc64)", but same compilation error)

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -15,6 +15,13 @@ prefix=joinpath(BinDeps.depsdir(libipopt),"usr")
 patchdir=BinDeps.depsdir(libipopt)
 srcdir = joinpath(BinDeps.depsdir(libipopt),"src",ipoptname)
 
+# fpu_control flag for building on ARM
+if Sys.ARCH == :arm
+    fpu_control = "ADD_CFLAGS=-DNO_fpu_control"
+else
+    fpu_control = ""
+end
+
 provides(SimpleBuild,
     (@build_steps begin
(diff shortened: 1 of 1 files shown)


mlubin:
Cool, does this actually work on ARM now?

jgonzal13:
Took a while to finish building, but yes `Pkg.test("Ipopt")` succeeded on an Odroid XU4 with this change.

mlubin:
LGTM, i'll let @tkelman sign off also

jgonzal13:
@tkelman did the typing here while at @jgonzal13's computer

mlubin:
That works then

joehuchette:
Neat!

jgonzal13:
We'd like to have this working easily for some other members of our lab, would it be okay to tag this quickly?

mlubin:
Will do

joehuchette:
https://github.com/JuliaLang/METADATA.jl/pull/3825

tkelman:
Just leaving a note that the ARM binaries of Julia are susceptible to `Provider BinDeps.BuildProcess failed to satisfy dependency libipopt` due to mismatched libstdc++ versions if Ipopt gets built by a newer version of `g++` than the ARM binaries were. Moving the copy of `lib/julia/libstdc++.so.6` that comes with in the Julia binaries out of the way seems to fix it.

PR #49, opened 2016-01-11
Tweak constructors to help with type inference

joehuchette:
Ref JuliaOpt/MathProgBase.jl/issues/94. Any reason not to do this @mlubin @IainNZ?


IainNZ:
Probably fine. Weird that it was explictly Any before, but, I guess that code is pretty old at this point. Tests pass so good enough for me.

mlubin:
LGTM

PR #52, opened 2016-02-17
Precompile some functions

joehuchette:


tkelman:
:+1:

Did you just follow the stock SnoopCompile readme instructions?

joehuchette:
Essentially, I ran the nonlinear benchmarks from https://github.com/mlubin/JuMPSupplement and went with that.

tkelman:
Should any of the `Int64`s actually be `Int` for the sake of 32 bit? Would that cause method errors or just not give the right speedup?

joehuchette:
Using ``Int`` is probably the right thing to do, but I'm pretty sure it doesn't cause any harm, just precompile an extraneous method.

odow:
Closing as stale. We should revisit precompile statements wholesale across JuMP-dev.

PR #55, opened 2016-07-12
Fix deprecations

abhijithch:
Fixes deprecations on `Julia v0.5`.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,4 +1,5 @@
 using BinDeps
+using Compat
 
 @BinDeps.setup
 
@@ -65,7 +66,7 @@ provides(SimpleBuild,
     end),libipopt, os = :Unix)
 
 # OS X
-@osx_only begin
+if is_apple()
     using Homebrew
     provides(Homebrew.HB, "ipopt", libipopt, os = :Darwin)
 end
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -1,6 +1,7 @@
 isdefined(Base, :__precompile__) && __precompile__()
 
 module Ipopt
+using Compat
 
 if isfile(joinpath(dirname(@__FILE__),"..","deps","deps.jl"))
     include("../deps/deps.jl")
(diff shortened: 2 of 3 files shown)


deps/build.jl
```diff
@@ -1,4 +1,5 @@
 using BinDeps
+using Compat
```
tkelman: set a minimum version in `REQUIRE` - you need to do this, every time

deps/build.jl
```diff
@@ -65,8 +66,8 @@ provides(SimpleBuild,
     end),libipopt, os = :Unix)
 
 # OS X
-@osx_only begin
-    using Homebrew
+@compat @compat if is_apple()    
+	using Homebrew
```
tkelman: spaces, not tabs

src/Ipopt.jl
```diff
@@ -204,7 +205,7 @@ function freeProblem(prob::IpoptProblem)
 end
 
 
-function addOption(prob::IpoptProblem, keyword::ASCIIString, value::ASCIIString)
+function addOption(prob::IpoptProblem, keyword::String, value::String)
```
tkelman: if the underlying library can't handle unicode, this needs to be checked that it's ascii

abhijithch: Can we use `Compat.ASCIIString` instead of `String`?

tkelman: that won't do any verification on 0.5 so doesn't address the issue

deps/build.jl
```diff
@@ -65,7 +66,7 @@ provides(SimpleBuild,
     end),libipopt, os = :Unix)
 
 # OS X
-@osx_only begin
+@compat @compat if is_apple()
```
tkelman: `@compat` not needed once, definitely not needed twice

src/Ipopt.jl
```diff
@@ -204,9 +205,12 @@ function freeProblem(prob::IpoptProblem)
...
     #/** Function for adding a string option.  Returns FALSE the option
     # *  could not be set (e.g., if keyword is unknown) */
+    if !isascii(keyword)
```
tkelman: and `value` ?

src/Ipopt.jl
```diff
@@ -240,7 +250,7 @@ function addOption(prob::IpoptProblem, keyword::ASCIIString, value::Integer)
 end
 
 
-function openOutputFile(prob::IpoptProblem, file_name::ASCIIString, print_level::Int)
+function openOutputFile(prob::IpoptProblem, file_name::String, print_level::Int)
```
tkelman: verification?

src/Ipopt.jl
abhijithch: Have updated the PR based on the comments. I think something is wrong with the `Julia` nightly build, issue posted [here](https://github.com/travis-ci/travis-ci/issues/6327).

mlubin:
@tkelman, I'll let you merge when this is ready

PR #56, opened 2016-07-22
Eliminate Homebrew ambiguity

staticfloat:
This full path to the formula is required when somebody has also installed something from Homebrew/science, thereby confusing `brew` between `homebrew/science/ipopt` and `staticfloat/juliadeps/ipopt`.


tkelman:
thanks!

PR #59, opened 2016-08-04
avoid deprecation warnings on 0.4

mlubin:


tkelman:
the homebrew postinstall failure is odd, but can't be related since you aren't touching build.jl here

PR #60, opened 2016-08-09
Fix @windows dep warn

ranjanan:
`@windows` -> `if is_windows()`


ranjanan:
@tkelman Could you tag a new release please?

PR #71, opened 2016-10-25
Don't eval `jac_structure` if problem has no constraints

abelsiqueira:


tkelman:
shouldn't you be implementing the empty jacobian in this case?

mlubin:
@tkelman, there's an argument that NLP evaluators for unconstrained models shouldn't need to do anything at all for jacobians

tkelman:
when trying to use a constrained solver?

mlubin:
MPB has no distinction between a constrained and unconstrained solver. I don't see the harm in not asking for jacobian information if we already know that the problem has no constraints.

mlubin:
@abelsiqueira, have you tested if Ipopt invokes the jacobian callbacks when there are no constraints?

abelsiqueira:
No, let me check.

abelsiqueira:
I tested it, it invokes the callback with `mode = :Structure`. What do you suggest I do?

mlubin:
@abelsiqueira, that's not an issue since we initialize ``Ijac, Jjac``.

dpo:
It's really IPOPT's job to allocate an vacuous Jacobian (if it really wants to) when a problem is unconstrained.

tkelman:
Not really, the solver API specifies what you're responsible for providing.

dpo:
It simply makes no sense to ask a user to implement `cons` and `jac` for an unconstrained problem.

tkelman:
then you're requiring separate api's for unconstrained vs constrained. from a primarily constrained solver's perspective, it makes perfect sense to only have the more general api.

dpo:
Nah.

mlubin:
@tkelman, I don't see the harm in accommodating this use case given that it doesn't complicate the code and @abelsiqueira has stepped forward to do the work.

If we want to make this more "formally" part of the MPB interface, it would be nice for the [unit tests](https://github.com/JuliaOpt/MathProgBase.jl/blob/master/test/nlp.jl) to have a case that covers an unconstrained problem where jac isn't implemented.

This also relates to the discussion at https://github.com/JuliaOpt/MathProgBase.jl/issues/87.

tkelman:
Are you going to require every solver to implement this the same way? Seems like you're getting lucky that Ipopt only evaluates this callback in structure mode here, but other constrained solvers may always require the same set of callbacks to be defined and always call them.

abelsiqueira:
I don't know who is calling the Jacobian callback, but perhaps it should not be called for an unconstrained problem. Otherwise, I can redefine the callback when the problem is unconstrained and solve locally.

I'm implementing this way because before it was always using the `:Jac` feature, regardless of availability.

For our models, we're considering prohibitive behaviour on calling functions not in a list of features.

mlubin:
@abelsiqueira, @tkelman's point is that you're now redefining the MPB interface. So let's do it correctly. We need a unit test at https://github.com/JuliaOpt/MathProgBase.jl/blob/master/test/nlp.jl for this behavior and the other open-source solvers (i.e., NLopt) need to be checked and updated as well.

tkelman:
I don't think MPB should be imposing that all constrained solvers have to support this.

mlubin:
@tkelman, it seems like a very small amount of work. Worst case you have a boolean flag that skips the call to ``eval_jac_g``.

tkelman:
Not if it's built into the way the solver works, several of which we won't be able to change.

mlubin:
@tkelman, what do you mean? This is only at the level at the MPB wrapper which we have full control over.

tkelman:
Aren't MPB wrappers supposed to be low level with as little required complication as possible on top of the solver API's themselves?

mlubin:
@tkelman, even so, replacing ``eval_jac_g(...)`` with ``m.has_jac && eval_jac_g(...)`` hardly seems to exceed the "little required complication" barrier.

mlubin:
(The translation from the triplet-with-duplicates form to upper-triangular-no-duplicates form [in KNITRO](https://github.com/JuliaOpt/KNITRO.jl/blob/67a6e8379d1b39476282abb56a6e65598e2896c6/src/KnitroSolverInterface.jl#L55) is more complicated, in fact.)

dpo:
If we can establish that Knitro attempts to evaluate the constraints/Jacobian on unconstrained problems, I'm happy to talk about it with its authors.

mlubin:
@abelsiqueira, if you're willing to add the unit tests to MathProgBase as discussed above, I'm okay proceeding with this.

abelsiqueira:
Okay, I'll look into it.

abelsiqueira:
@mlubin , it looks like `nlp.jl` is not being used ATM. There's no `include` to it.

mlubin:
@abelsiqueira, the includes are at the solvers: https://github.com/JuliaOpt/Ipopt.jl/blob/6a5515460921c2fa27d00a668bb232a1ad95aa0c/test/runtests.jl#L54
https://github.com/JuliaOpt/KNITRO.jl/blob/9b1fe1279709b6de2abebb2026c76d8829a3c23a/test/mathprogtest.jl#L1

Issue #72, opened 2016-11-26
build fails on travis


mlubin:
https://travis-ci.org/JuliaOpt/JuMP.jl/jobs/178999979#L3880

```
Downloading the source code from projects.coin-or.org...
--2016-11-26 06:18:56--  http://www.coin-or.org/BuildTools/Blas/blas-20130815.tgz
Resolving www.coin-or.org (www.coin-or.org)... 130.127.206.21
Connecting to www.coin-or.org (www.coin-or.org)|130.127.206.21|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://www.coin-or.org/BuildTools/Blas/blas-20130815.tgz [following]
--2016-11-26 06:18:56--  https://www.coin-or.org/BuildTools/Blas/blas-20130815.tgz
Connecting to www.coin-or.org (www.coin-or.org)|130.127.206.21|:443... connected.
ERROR: cannot verify www.coin-or.org's certificate, issued by `/C=US/ST=Arizona/L=Scottsdale/O=GoDaddy.com, Inc./OU=http://certs.godaddy.com/repository//CN=Go Daddy Secure Certificate Authority - G2':
  Unable to locally verify the issuer's authority.
To connect to www.coin-or.org insecurely, use `--no-check-certificate'.
Failed, try downloading the source code from www.netlib.org...
--2016-11-26 06:18:56--  http://www.netlib.org/blas/blas.tgz
Resolving www.netlib.org (www.netlib.org)... 160.36.131.221
Connecting to www.netlib.org (www.netlib.org)|160.36.131.221|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 101112 (99K) [application/x-gzip]
Saving to: `blas.tgz'
100%[======================================>] 101,112     --.-K/s   in 0.06s   
2016-11-26 06:18:57 (1.51 MB/s) - `blas.tgz' saved [101112/101112]
Uncompressing the tarball...
Unpacking the source code...
Deleting the tar file...
Moving the source files from BLAS subdirectory
mv: cannot stat `BLAS/*.f': No such file or directory
================================[ ERROR: Ipopt ]================================
LoadError: failed process: Process(`./get.Blas`, ProcessExited(1)) [1]
while loading /home/travis/.julia/v0.5/Ipopt/deps/build.jl, in expression starting on line 81
```

mlubin:
Fixed by #73

PR #73, opened 2016-11-26
work around for travis failing due to certificate issues (#72)

mlubin:
CC @tkelman

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
     - osx
 julia:
     - 0.4
+    - 0.5
     - nightly
 notifications:
     email: false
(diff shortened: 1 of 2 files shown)


tkelman:
What changed here? Should probably tell someone if the certificate needs updating. Or if it's a Travis only problem, maybe coin-or's cert is newer than what ubuntu 12.04 recognizes, and either switching to trusty or updating ca-certificates might help. This patch insecure and not necessarily a good idea.

tkelman:
This also isn't going to work on mac if anyone needs to fall back to source build there.

mlubin:
@tkelman, the patch is not insecure because we have been using insecure ``http`` this whole time. The ``http`` addresses seem to be redirecting to ``https`` now which doesn't work on 12.04. I'll try playing around with getting travis working another way.

PR #74, opened 2016-11-27
try another way

mlubin:
Ref https://github.com/JuliaOpt/Ipopt.jl/pull/73

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,13 @@ julia:
     - nightly
 notifications:
     email: false
-sudo: false
-addons:
-    apt_packages:
-        - gfortran
+dist: trusty
+sudo: required
+#addons:
+#    apt_packages:
+#        - gfortran
 script:
+    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -y install gfortran ; fi
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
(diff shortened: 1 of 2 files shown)


mlubin:
@tkelman, this is a less hacky approach but means that we have to make these travis changes for all packages that build ipopt in their tests.

mlubin:
(Accidentally deleted my comment while fighting with @coveralls)
@tkelman this is a less hacky approach but means that we have to use these travis settings on all packages that build Ipopt in their tests.

mlubin:
CC @ccoffrin

tkelman:
how about apt-get upgrading ca-certificates?

mlubin:
@tkelman, can you do that without sudo?

tkelman:
It might be in the apt packages plugin whitelist, would have to check. Maybe just a mac-compatible version of the sed call would be simpler.

mlubin:
@tkelman, I tried listing ``ca-certificates`` under ``apt_packages``, didn't fix anything.

ccoffrin:
I have been using,

```
addons:
  apt_packages:
    - gfortran
    - liblapack-dev
    - libgmp-dev
    - libglpk-dev
```

Does this resolve the issue?

mlubin:
Try reruning a travis build. I'd bet Ipopt won't build correctly with that
configuration anymore.

ccoffrin:
You are correct, https://github.com/ccoffrin/j5-ipopt-test

mlubin:
Closed by #75

PR #75, opened 2016-11-29
build script fix for OS X

mlubin:
@tkelman @ccoffrin, this should fix the source build on OS X and doesn't require any travis changes for upstream packages.


ccoffrin:
That is black magic to me.

PR #76, opened 2016-12-10
drop support for 0.4

mlubin:


tkelman:
Bad timing. Please release a version of the package that passes tests on 0.4 first. http://pkg.julialang.org/logs/Ipopt_0.4.log

PR #85, opened 2017-06-22
adding mpb interface setparameters! 

jac0320:
make the slience/timelimit work universally

### DIFF
--- a/src/IpoptSolverInterface.jl
+++ b/src/IpoptSolverInterface.jl
@@ -115,7 +115,7 @@ function setparameters!(s::IpoptSolver; mpboptions...)
             push!(opts, (:max_cpu_time,optval))
         elseif optname == :Silent
             if optval == true
-                push!(opts, (:log_level,0))
+                push!(opts, (:print_level,0))
             end
         else
             error("Unrecognized parameter $optname")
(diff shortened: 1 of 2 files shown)


tkelman:
Looks reasonable. Tests would be ideal.

mlubin:
Here's an example test from Cbc: https://github.com/JuliaOpt/Cbc.jl/blob/59b815c07aa4165112743b6ebdbfc7095dfc8728/test/runtests.jl#L10

jac0320:
* Modified one method of setparameters! given there are cases m.inner is not loaded
* Added tests

jac0320:
The Ipopt `brand log` default cannot be silent. But as long as Ipopt is not restarted through command line (like AmplNLWriter->Bonmin calls Ipopt), the brand log will only shows up once.

natschil:
This seems like it would have been a really useful pull-request, I'm now calling `push!(ipsol.options, (:print_level, 0))
` like it does as setparameters! is not available. However, as the pull-request was closed I'm now worried this could cause something to go wrong, what is the reason for closing the pull request?

odow:
This is no longer relevant.

Use 
```Julia
set_silent(model)
# or
set_optimizer_attribute(model, "print_level", 0)
```

PR #92, opened 2017-11-12
Fix depwarns on Julia 0.7 (bumps REQUIRE to 0.6)

timholy:


mlubin:
Could you remove 0.5 on travis?

mlubin:
Let's wait for https://github.com/JuliaLang/METADATA.jl/pull/11965 to go through, then bump MPB requirements here to get tests passing.

mlubin:
MPB 0.7 is tagged, so just need to bump the upper bound to 0.8 here.

timholy:
Seems like this is all set.

PR #96, opened 2018-01-10
Compile a newer Ipopt version

timholy:
I encountered what appears to be an [Ipopt bug](https://projects.coin-or.org/Ipopt/ticket/297) and had some hope it would be fixed by the latest version (no such luck).


mlubin:
I'll ignore the failure on OS X nightly since it likely wasn't caused by this.

PR #101, opened 2018-03-02
Note on supported configurations

mlubin:
This note reflects the current design of BinaryBuilder.

There will be linux users who previously had a working install using the BinDeps build script who will now need to manually compile from source, or switch to using the official binaries.

CC @staticfloat @stefankarpinski @tkelman

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,13 @@ platform is not supported, or if you prefer to compile your own version of Ipopt
 in order to use commercial sparse linear algebra libraries, use the instructions
 below.
 
+Note that currently the only supported configuration for the
+default installation is to use the
+[official Julia binaries](https://julialang.org/downloads/). If you build
+Julia from source or install Julia through a package manager, you may need to
+follow the custom installation instructions below.
+
+
 **Custom Installation**:
 
 Make sure you have the required packages before installing, e.g.,


tkelman:
This applies only to master for now, and is a bad enough user experience that I'm asking you not to release this package in this state. If BinaryBuilder is a strict improvement on Mac, go ahead and use it there. But as long as it has this major flaw on Linux I don't think it should be used there yet.

tkelman:
If the build configuration of JuliaPro hasn't changed much since I last looked at it, I'm fairly sure that master of this package won't work there, for example.

mlubin:
> But as long as it has this major flaw on Linux I don't think it should be used there yet.

My understanding is that this is Julia Computing's design for how binaries will be handled on Linux for all packages. There's not much of a point for Ipopt.jl to bend over backwards to support use cases that the rest of the package ecosystem won't support. I hope others will chime in on this point.

> If the build configuration of JuliaPro hasn't changed much since I last looked at it, I'm fairly sure that master of this package won't work there, for example.

Julia Computing supports both BinaryBuilder and JuliaPro, so if there's indeed an issue I would expect it to be resolved on their side. This isn't a responsibility that should fall on the open-source maintainer of Ipopt.jl.

tkelman:
I don't think that's correct that this is intended to be a permanent design decision, it's an early limitation because BinaryBuilder isn't complete yet and doesn't satisfy a lot of use cases that it will really need to before it can actually be effectively used. I don't think this limitation is well understood or communicated in BinaryBuilder.

tkelman:
JuliaComputing should definitely be doing maintenance required for JuliaPro, but what I'm saying is I don't think they're fully aware that BinaryBuilder in its current state produces binaries that will often not be usable for source builds of Julia (no-GPL builds, MKL builds, etc). If this were widely known and established, I highly doubt anyone would be advertising BinaryBuilder as ready for widespread use. Julia Computing doesn't want to have to patch and maintain custom builds for every package they distribute.

mlubin:
@tkelman, the JC folks are willing to address these issues as they arise. I understand your objections but I disagree on priorities. I tested that the build script continues to work when upgrading from a BinDeps source build. Users who do a fresh install of Ipopt on a setup that's not supported by the new binaries will indeed be inconvenienced. I would support a build-from-source fallback in build.jl but it's not a sustainable use of my time to implement a hacky fallback that introduces a complex dependency (BinDeps). A couple things that would need to be resolved are:

- The deps.jl format is a bit different between BinDeps and BinaryProvider.
- The source build needs to be tested in CI.

It would be better for this work to come from someone who has a good motivation to have these minority use cases running smoothly. Given that the current release has a major defect on OS X, I intend to tag a new release this week unless there are additional objections that have not been considered yet.

tkelman:
It's not introducing a new dependency, it's putting the old code back behind an if statement.

mlubin:
> It's not introducing a new dependency, it's putting the old code back behind an if statement.

I would consider a PR that does this and addresses the two points I raised above. However, I'm also recognizing the additional maintenance costs of depending on an obscure build system that few people will be able to fix when something goes wrong. For legacy reasons I'm willing to put those concerns aside, but I'm not willing to be the owner of this code.

mlubin:
Replaced by #103

PR #103, opened 2018-03-05
Restore BinDeps use on non-Mac, non-Windows platforms

tkelman:
if the BinaryProvider artifacts assume libgfortran.so.4 is present,
they will not work for any source builds of Julia itself when a GCC
version older than 7 was used

as discussed in https://github.com/JuliaOpt/Ipopt.jl/pull/101
best viewed as https://github.com/JuliaOpt/Ipopt.jl/pull/103/files?w=1

diff w.r.t. pre-BinaryProvider master: https://github.com/JuliaOpt/Ipopt.jl/compare/6de0898ba2797fbba6bd6cfced89a4e2154130cf...tk/linux-bindeps?w=1


deps/build.jl
```diff
@@ -1,40 +1,135 @@
...
+                            --with-lapack=$prefix/lib/libcoinlapack.a
+                            $fpu_control`
+                `make`
+                `make test`
+                `make -j1 install`
+            end
+        end),libipopt, os = :Unix)
+
+    # Windows
+    if is_windows()
```
mlubin: What's the motivation for reverting to WinRPM on Windows?

src/Ipopt.jl
```diff
@@ -8,14 +8,17 @@ if isfile(joinpath(dirname(@__FILE__),"..","deps","deps.jl"))
 else
     error("Ipopt not properly installed. Please run Pkg.build(\"Ipopt\")")
 end
+if !use_BinaryProvider
```
mlubin: Add a comment that `use_BinaryProvider` is defined in deps.jl.

deps/build.jl
```diff
@@ -1,40 +1,135 @@
...
+        provides(WinRPM.RPM, "Ipopt", [libipopt], os = :Windows)
+    end
+
+    # Avoid Issue #62, building into /lib64 on OpenSUSE
+    configsite = nothing
+    if haskey(ENV, "CONFIG_SITE")
+        configsite = pop!(ENV, "CONFIG_SITE")
+    end
+
+    @BinDeps.install Dict(:libipopt => :libipopt)
```
mlubin: BinDeps will pick up a system install of ipopt, right? This is inconsistent with what's written in the README and with what now happens on macOS.

deps/build.jl
tkelman: could go either way, was restoring the parts that were known to work and only removing that parts that are known not to

deps/build.jl
```diff
@@ -1,40 +1,135 @@
...
+        fpu_control = "ADD_CFLAGS=-DNO_fpu_control"
     else
-        # If we don't have a BinaryProvider-compatible .tar.gz to download, complain.
-        # Alternatively, you could attempt to install from a separate provider,
-        # build from source or something more even more ambitious here.
-        error("Your platform $(Sys.MACHINE) is not supported by this package!")
+        fpu_control = ""
+    end
+
+    provides(SimpleBuild,
```
mlubin: Suppose the user has tried to build from source to use a different linear algebra library but has a slightly broken install, e.g., that doesn't `dlopen`. In this case will the source build trigger and attempt to overwrite the install?

tkelman: I believe so

deps/build.jl
tkelman: believe so, though the validation check here isn't very strict to detect if it'll actually work (most distro copies of ipopt have historically been broken, though that might have changed on some distros by now)

deps/build.jl
mlubin: The new binaries appear to work on Windows, so I don't see a reason not to try and use them. Very few people know how to maintain the WinRPM scripts.

deps/build.jl
mlubin: To be fair there's a similar issue with the BinaryProvider script

deps/build.jl
tkelman: I can change it over, but how many know how to maintain the binarybuilder scripts?

mlubin: An increasing number since the plan is to transition all JuliaOpt solver binaries over to BinaryBuilder. BinaryBuilder is also supported by Julia Computing while WinRPM is on its way out.

deps/build.jl
mlubin: How difficult is it to stop picking up system installs? We don't want people building workflows around custom builds by modifying `LD_LIBRARY_PATH` only for this to be broken later.

deps/build.jl
```diff
@@ -1,40 +1,129 @@
...
-    BinaryProvider.Windows(:x86_64) => ("$bin_prefix/Ipopt.x86_64-w64-mingw32.tar.gz", "d967b152449659591095458594948f74981fc0b772cbf3f333a9a24676beab80"),
-)
-
-# First, check to see if we're all satisfied
-if any(!satisfied(p; verbose=verbose) for p in products)
-    if haskey(download_info, platform_key())
-        # Download and install binaries
-        url, tarball_hash = download_info[platform_key()]
-        install(url, tarball_hash; prefix=prefix, force=true, verbose=verbose)
+use_BinaryProvider = Compat.Sys.isapple()
```
mlubin: Missing check for windows

deps/build.jl
tkelman: Do you have any way of testing that? We could try to temporarily remove SystemPaths from BinDeps list of default providers but I'm not sure whether that would totally work.

deps/build.jl
tkelman: duh, fixed

deps/build.jl
mlubin: The debian package `coinor-libipopt-dev` is likely detected by default.

tkelman: right but this wouldn't be a regression relative to the last release. if I had to guess, I suspect far more people use source builds of julia than have broken installs of system ipopt packages. if the system ipopt works, using it is fine. if it's broken, why keep it installed? I doubt anything else in debian depends on it. we can do better validation separately as a fix for an already-broken combination. I think telling broken from working system ipopt is a little easier than distinguishing source build from binary install of julia, but BinDeps specific validation would need to be partially rewritten whenever BinaryBuilder finishes supporting all the linux use cases that BinDeps has worked for.

mlubin:
> whenever BinaryBuilder finishes supporting all the linux use cases that BinDeps has worked for

I don't think there are any plans to do this.

Anyway, please document:
- The library search path behavior under all configurations (the README is now incorrect). This can either be all in the README or split across the README and `build.jl`.
- The motivations for choosing to not use the BinaryBuilder linux binaries (i.e., as a comment in `build.jl`)

tkelman:
> I don't think there are any plans to do this.

Perhaps not all, but aside from interaction with system package managers, what situations where BinDeps did work are you under the impression BinaryBuilder is explicitly intending to not work, by design, indefinitely?

deps/build.jl
tkelman: You keep saying "BinaryBuilder is also supported by Julia Computing" but so far there has been zero input from either its primary developer who to my knowledge isn't affiliated with Julia Computing, or anyone else who is, about how they plan to address shortcomings that fail to support a common use case that even they require to build their own products.

mlubin:
Discussion on BinaryBuilder is taking place on the bindeps2 channel on the public julia slack. It's not efficient for me to continue being the intermediary, so that's the best place to ask about the configurations that they intend to support or not.

tkelman:
which is neither permanent nor reflected in documentation

mlubin:
I'm well aware of that but not in a position to address it.

README.md
```diff
@@ -13,8 +13,9 @@ Ipopt.jl
 
 This will install Ipopt.jl, as well as Ipopt itself. A binary will be downloaded
 by default on macOS or Windows, and Ipopt will be automatically built from source
-on Linux. If your platform is not supported, or if you prefer to compile your own
-version of Ipopt in order to use commercial sparse linear algebra libraries, use
+on Linux unless a pre-existing version is found on the `LD_LIBRARY_PATH`.
+If your platform is not supported, or if you prefer to compile your own version
+of Ipopt in order to use commercial sparse linear algebra libraries, use
```
mlubin: There's also a statement about the search path that needs to be updated under the custom installation section.

Issue #110, opened 2018-04-12
Ipopt build error on Mac


gdoshi9:
I am trying to add and build Ipopt on version 0.6.2 of Julia on Mac. It does install (but incorrectly) and when I try to build it, I get the following error:
```
INFO: Building Ipopt
WARNING: imported binding for srcdir overwritten in module __anon__
================================[ ERROR: Ipopt ]================================

LoadError: MethodError: no method matching BinaryProvider.LibraryProduct(::BinaryProvider.Prefix, ::Array{String,1}, ::Symbol)
Closest candidates are:
  BinaryProvider.LibraryProduct(::BinaryProvider.Prefix, ::Array{S<:AbstractString,1}) where S<:AbstractString at /Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/BinaryProvider/src/Products.jl:42
  BinaryProvider.LibraryProduct(::BinaryProvider.Prefix, ::AbstractString) at /Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/BinaryProvider/src/Products.jl:38
  BinaryProvider.LibraryProduct(::AbstractString, ::Array{S<:AbstractString,1}) where S<:AbstractString at /Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/BinaryProvider/src/Products.jl:56
while loading /Applications/JuliaPro-0.6.2.1.app/Contents/Resources/pkgs-0.6.2.1/v0.6/Ipopt/deps/build.jl, in expression starting on line 13

================================================================================
UndefVarError: BinaryProvider not defined
```
I saw that someone else is facing the same issue on windows as well.

mlubin:
@staticfloat, was there a breaking change in BinaryProvider?

mlubin:
Oh, ok. Then we at least need a lower bound in REQUIRE.

GauravDoshi9:
Actually this is the first time I have come across BinaryProvider, so I don't have any idea regarding the version of BinaryProvider or its update. But doesn't it get updated when we run Pkg.update()?

staticfloat:
Yes, it should.  Most likely it was installed by some other package in the past, and you just haven't run `Pkg.update()` in a while?  Try running `Pkg.update()` and try again.

GauravDoshi9:
[...]

PR #111, opened 2018-04-12
Lower bound on BinaryProvider dependency

mlubin:


tkelman:
Will need to make the change in the metadata copy of the existing tag for this to make much difference

REQUIRE
```diff
@@ -1,5 +1,5 @@
 julia 0.6
 MathProgBase 0.5 0.8
-BinaryProvider
+BinaryProvider 0.2.5
```
tkelman: isn't the minimum 0.2.0 for the problem from #110?

mlubin: Lacking further information I picked the version of BinaryProvider that was available when we last tagged Ipopt. I don't feel a need to spend the time to test it on earlier releases.

tkelman: it's not good to bound things unnecessary tightly, at least look for the specific api change in question

mlubin: Sure, but there's also a risk of that the loose version bounds don't work, e.g., https://github.com/JuliaOpt/Ipopt.jl/issues/105. Given that BinaryBuilder is changing quickly, there's a real possibility of this happening if we allow versions that were never tested.

tkelman: and if that causes a real problem it can be tightened to fix it after it gets reported and the correct bound is identified. nothing since 0.2.0 looks like it would make a major difference here across all platforms.

tkelman: People who have it pinned, or are using JuliaPro if that happens to be the release. You're preventing installation of version combinations that would work, which is worse than allowing combinations that should work but haven't been tested. Lower bounds aren't for "has been tested," they are for "lowest that should be allowed at all." It's correctness, not style.

mlubin: Your response doesn't address my second question of why we should be putting effort into serving, 

> People who have it pinned, or are using JuliaPro if that happens to be the release.

> You're preventing installation of version combinations that would work, which is worse than allowing combinations that should work but haven't been tested.

This is a software engineering style judgement and not a correctness issue. As maintainers of the package we choose which versions of the dependencies we intend to support. If the version is allowed but doesn't actually work, that's one more issue that will be opened on the repository, which has its own maintenance costs.

tkelman: It's taking more time to argue the point than it takes to look at the history of the source code of the dependency in question and identify the actual change this PR is intended to account for.

When you disallow such a combination that would work fine, there's similarly an issue that would be opened by such a person who needs to support an environment that has a slightly older version of BinaryProvider pinned due to the requirements of other packages in that environment. And it's a one-character change to avoid that.

mlubin: You're right, I tested 0.2 and it works.

PR #113, opened 2018-04-29
Updates for latest MOI

odow:
Remove dependence on MathOptInterfaceBridges now that it has been moved to MOI. 

Locally I'm getting a weird error. All the tests pass, an no error is thrown, but I get test errors...

From Travis, [it is a problem with the finalizer](https://travis-ci.org/JuliaOpt/Ipopt.jl/jobs/372586912#L4520-L4529).


test/runtests.jl
```diff
@@ -1,82 +1,37 @@
...
+
+    # Test that the ipopt binary works
+    @test success(`$(Ipopt.amplexe) -v`)
+end
+
+@testset "MathProgBase" begin
+    include("MPBWrapper.jl")
+end
+
+@testset "MathOptInterface" begin
```
blegat: Should we add an advice to do this in the guideline ?

test/runtests.jl
```diff
@@ -21,7 +21,6 @@ using Ipopt, Base.Test
     # Test opening an output file
     openOutputFile(prob, "blah.txt", 5)
     # unlink the output file
-    Ipopt.freeProblem(prob)
```
odow: I think I had to add that in order to free the lock on the output file.

mlubin: Well that's no good because it calls `freeProblem` twice.

mlubin: We can just add some logic in `freeProblem` to avoid this. But that can be a separate PR.

PR #120, opened 2018-06-16
Set PATH so that Ipopt binary works on Windows.

### LINKED ISSUE #119: Ipopt.exe doesn't run on Windows/Julia 0.7
```
C API: Test Failed at C:\Users\Miles\.julia\dev\Ipopt\test\runtests.jl:28
  Expression: success(#= C:\Users\Miles\.julia\dev\Ipopt\test\runtests.jl:28 =# @cmd("\$(Ipopt.amplexe) -v"))
```

Recall that the exe has weird dependencies on the libraries that are only available when running Julia: https://github.com/JuliaOpt/Ipopt.jl/blob/3da72b6f51a78bb0daf049a9bef12c57944841a0/src/Ipopt.jl#L23

So something along these lines probably changed between 0.6 and 0.7

@staticfloat

mlubin:
Fixes #119 
Thanks @staticfloat!
The remaining failure on 0.7 is because https://github.com/JuliaOpt/MathOptInterface.jl/commit/48059841418f92017413b5ecf1bebc02d2026568 isn't tagged.

### DIFF
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -2,6 +2,7 @@ __precompile__()
 
 module Ipopt
 using Compat
+import Compat.Libdl
 using Compat.LinearAlgebra
 
 if isfile(joinpath(dirname(@__FILE__),"..","deps","deps.jl"))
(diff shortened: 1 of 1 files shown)


tkelman:
This really needs to be in a withenv so it's temporary. Permanently adding Julia's bin to path is going to break other applications.

mlubin:
There's a TODO in the code to address this. The fix involves changing all downstream uses of ipopt binary (e.g., in AmplNLWriter). What is in `Sys.BINDIR` that's particularly liable to break other applications if it appears in the path?

tkelman:
The mingw runtimes, precisely what you're looking for here. Julia isn't the only application or library out there built with mingw, and there are incompatible versions or ways in which they can be configured.

Yes this is the wrong place to be making this change - the callers should be responsible for this.

mlubin:
How does the Windows dll resolution process work? Does putting one directory with mingw runtimes in the PATH break all other mingw applications build on other runtimes? That seems unusual.

tkelman:
I have seen that happen, yes. Julia itself used to do this in the default windows juliarc but it wasn't actually necessary or beneficial in that case.

mlubin:
Ok, good to know. I guess this fix will hang around for a while until someone is motivated to refactor how AmplNLWriter calls binaries.

mlubin:
This is stale

PR #121, opened 2018-06-24
disable executable test until #119 is fixed

mlubin:


tkelman:
Why not do the withenv here?

mlubin:
Just for 0.7/Windows? We could, but it's unusual to test that without having any public API for it.

tkelman:
Well the tests here won't exercise that the build of ipopt.exe works, but since it's provided here mostly as a courtesy for AmplNLWriter maybe you don't care

PR #130, opened 2018-08-10
More 0.7 updates, drop 0.6

alyst:


mlubin:
@alyst Thanks! Please note that we're not ready to drop support for Julia 0.6 on solver packages (https://discourse.julialang.org/t/jump-and-julia-0-7-1-0/12964). Could you maintain compatibility using `Compat`?

PR #133, opened 2018-08-15
better handling of NaN/Inf in callbacks

### LINKED ISSUE #132: Ipopt fails on simple constr. opt. problem
The repo at https://github.com/arbenede/error_reports.git contains contains some code to solve a simple nonlinearly constrained optimization problems that fails under both Windows and Linux (Windows log is in the repo). The Linux error file was posted a few days ago by @tkoolen at https://gist.github.com/tkoolen/79071106c93e3c974cdb11f5a44269df. Apparently it works under OSX (see the thread at https://discourse.julialang.org/t/jump-ipopt-failing-on-simple-problem/13209 )

mlubin:
Closes #132 

@arbenede @tkoolen

### DIFF
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -674,6 +674,12 @@ function MOI.optimize!(m::IpoptOptimizer)
                             length(hessian_sparsity),
                             eval_f_cb, eval_g_cb, eval_grad_f_cb, eval_jac_g_cb,
                             eval_h_cb)
+
+    # Ipopt crashes by default if NaN/Inf values are returned from the
+    # evaluation callbacks. This option tells Ipopt to explicitly check for them
+    # and return Invalid_Number_Detected instead.
+    addOption(m.inner, "check_derivatives_for_naninf", "yes")
+
     if !has_hessian
         addOption(m.inner, "hessian_approximation", "limited-memory")
     end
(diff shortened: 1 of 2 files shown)


src/MOIWrapper.jl
```diff
@@ -674,6 +674,12 @@ function MOI.optimize!(m::IpoptOptimizer)
                             length(hessian_sparsity),
                             eval_f_cb, eval_g_cb, eval_grad_f_cb, eval_jac_g_cb,
                             eval_h_cb)
+
+    # Ipopt crashes by default if NaN/Inf values are returned from the
+    # evaluation callbacks. This option tells Ipopt to explicitly check for them
+    # and return Invalid_Number_Detected instead.
+    addOption(m.inner, "check_derivatives_for_naninf", "yes")
```
tkelman: huh. I was under the impression this was on by default for some reason

tkelman: ah not for derivatives, but maybe I've seen it check objective and constraint primal return values by default since there aren't as many of those values

tkelman: I do kinda hate when modeling languages and/or solver wrappers unconditionally overwrite options with their own different defaults though - crashing isn't friendly but we should let people who want better performance and know they have well-behaved functions turn this back off manually.

mlubin: I added a more useful comment for anyone who is digging through the code.

tkelman:
If you call addOption multiple times for the same option does it use the value from the first call or the last call?
________________________________

mlubin:
`addOption` translates directly to a call to the Ipopt API, so it's hard to imagine that if you call it multiple times it won't use the last value set.

mlubin:
I do not know for sure. You can check if this is an important use case for
you. I can prioritize fixing crashes but there are other more much
significant performance issues to address in JuMP/MOI at the moment.

tkelman:
It's important for solver wrappers to avoid setting options in a non-overridable way. If the underlying API uses the first value set for an option instead of the last, then this kind of hard-coding should only be done if the user-passed options don't include a value for the corresponding option.

PR #147, opened 2018-10-20
store options as vector of tuples instead of Iterator of pairs

Wikunia:
This is done to be able to change the options inside other solvers by changing the vector of tuples as it was possible in julia v0.6.
It shouldn't change any behavior of Ipopt.

Reference: @mlubin 
https://discourse.julialang.org/t/updating-options-in-ipopt/16013/2

### DIFF
--- a/src/MPBWrapper.jl
+++ b/src/MPBWrapper.jl
@@ -5,7 +5,7 @@ const MPB = MathProgBase
 # Solver objects
 export IpoptSolver
 struct IpoptSolver <: MPB.AbstractMathProgSolver
-    options::Vector{Tuple}
+    options::Vector{Tuple} # list of options set in Ipopt on each MPB.optimize! call
 end
 function IpoptSolver(;kwargs...)
     args = Vector{Tuple}()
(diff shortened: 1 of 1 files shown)


src/MPBWrapper.jl
```diff
@@ -7,7 +7,17 @@ export IpoptSolver
 struct IpoptSolver <: MPB.AbstractMathProgSolver
     options
```
mlubin: Please change this to `options::Vector{Tuple}` to make the intention of the code explicit. It would also be nice to have some documentation about what happens when you manually change the `options` vector between solves, because this was never a documented API.

Wikunia: Can you clarify the part about the documentation a little bit?
- Where can I document it best? Is there a documentation of Ipopt.jl?
- The options vector is read every time `JuMP.solve` is called as this calls `MPB.optimize!` which gets the options set in the internal representation and then pushes the options to the C interface.

Wikunia:
Alright thanks. Traivs seemed to have a small problem with apple. Probably just needs a restart.

mlubin:
BinDeps crashing on OS X has been a long-term issue.

PR #157, opened 2019-01-21
test callback

dpo:
The callback functionality is not currently being tested. A callback was defined but not registered with the problem. On macOS, the test fails with the message:
```
C API: Error During Test at /Users/dpo/dev/julia/Ipopt.jl/test/runtests.jl:5
  Got exception outside of a @test
  LoadError: ccall: could not find function SetIntermediateCallback in library /Users/dpo/dev/julia/Ipopt.jl/deps/usr/lib/libipopt.1.10.10.dylib
```
The issue is that
```
dpo:~/dev/julia/Ipopt.jl (test-callback %)$ nm /Users/dpo/dev/julia/Ipopt.jl/deps/usr/lib/libipopt.1.10.10.dylib | grep -i SetIntermediateCallback
00000000000151b0 t _SetIntermediateCallback
```
The problem presumably comes from IpoptBuilder.jl.

### DIFF
--- a/test/hs071_test.jl
+++ b/test/hs071_test.jl
@@ -96,14 +96,6 @@ function eval_h(x::Vector{Float64}, mode, rows::Vector{Int32}, cols::Vector{Int3
   end
 end
 
-function intermediate(alg_mod::Int, iter_count::Int,
-  obj_value::Float64, inf_pr::Float64, inf_du::Float64, mu::Float64,
-  d_norm::Float64, regularization_size::Float64, alpha_du::Float64, alpha_pr::Float64,
-  ls_trials::Int)
-  println("Iteration $iter_count, objective value is $obj_value.")
-  return true
-end
-
 n = 4
 x_L = [1.0, 1.0, 1.0, 1.0]
 x_U = [5.0, 5.0, 5.0, 5.0]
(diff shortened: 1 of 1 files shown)


dpo:
cf https://github.com/JuliaOpt/IpoptBuilder/issues/2

test/hs071_test.jl
```diff
@@ -116,6 +116,7 @@ prob = createProblem(n, x_L, x_U, m, g_L, g_U, 8, 10,
                      eval_f, eval_g, eval_grad_f, eval_jac_g, eval_h)
 
 prob.x = [1.0, 5.0, 5.0, 1.0]
+setIntermediateCallback(prob, intermediate)
```
mlubin: Instead of printing, the callback should set a flag so that we can test that the callback was called. (I realize this is a pre-existing issue with the test.)

dpo: Sure but the more pressing issue is that we're not able to set a callback at all.

mlubin: Obviously. :)
This is a comment to be addressed after callbacks are working again.

dpo: There we go.

juan-pablo-vielma:
added #161 with the updated binaries that should allow this PR to pass

dpo:
Thank you!

Issue #158, opened 2019-01-30
optimization randomly results in INVALID_MODEL (possible initialization bug?)


pseudocubic:
I'm having a bit of a weird problem with non-deterministic optimization failures and I'm not really sure where the problem is. I originally encountered this in context of using PowerModels, but I've been able to boil it down to the demonstration code below. In Julia 1.0.3 everything works fine, but in Julia 1.1 the `JuMP.termination_status` obtained after `JuMP.optimize!` can occasionally (at random) result in `MathOptInterface.INVALID_MODEL`:

```
using JuMP, Ipopt
m = for i in 1:10000
    m = JuMP.Model()
    a = JuMP.@variable(m, [j in 1:3], start=0.0)
    JuMP.@constraint(m, -1 .<= a .<= 1)
    ipopt_optimizer = JuMP.with_optimizer(Ipopt.Optimizer, print_level=0)
    JuMP.optimize!(m, ipopt_optimizer)
    ts = JuMP.termination_status(m)
    if ts != MOI.LOCALLY_SOLVED
        println("$i failed: $ts")
    end
end
```

I am running MacOS 10.14.2 with Julia 1.0.3, and Julia 1.1.0 (with the same environment in each), JuMP master, MOI 0.8.1, Ipopt 0.5.2.

I'm a bit at a loss what would have changed between Julia 1.0.3 and Julia 1.1 that would have caused these random failures.

@mlubin indicated that this may be a possible initialization bug. Below is some of the Ipopt output for cases where the termination status is INVALID_MODEL:

With `print_level=3`:
```
Total number of variables............................:        3
                     variables with only lower bounds:        0
                variables with lower and upper bounds:        0
                     variables with only upper bounds:        0
Total number of equality constraints.................:        0
Total number of inequality constraints...............:        6
        inequality constraints with only lower bounds:        3
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        3


Number of Iterations....: 3

Number of objective function evaluations             = 4
Number of objective gradient evaluations             = 4
Number of equality constraint evaluations            = 0
Number of inequality constraint evaluations          = 4
Number of equality constraint Jacobian evaluations   = 0
Number of inequality constraint Jacobian evaluations = 1
Number of Lagrangian Hessian evaluations             = 1
Total CPU secs in IPOPT (w/o function evaluations)   =      0.001
Total CPU secs in NLP function evaluations           =      0.000

[...]

PR #159, opened 2019-01-31
Fix gradient bug when no objective set

odow:
I couldn't get the failure reported in #158 to trigger, but this seems the likely culprit.

@pseudocubic, can you test please?


pseudocubic:
It looks like this fixed it, thanks!

odow:
I've created an issue: https://github.com/JuliaOpt/MathOptInterface.jl/issues/653

PR #165, opened 2019-05-23
Add SolverName Attribute to MOI Interface

ccoffrin:


test/MOI_wrapper.jl
```diff
@@ -20,6 +20,10 @@ const optimizer = Ipopt.Optimizer(print_level=0, fixed_variable_treatment="make_
 const config = MOIT.TestConfig(atol=1e-4, rtol=1e-4,
                                optimal_status=MOI.LOCALLY_SOLVED)
 
+@testset "MOI API tests" begin
+  @test MOI.get(optimizer, MOI.SolverName()) == "Ipopt"
```
blegat: 4-space indent

test/MOI_wrapper.jl
```diff
@@ -20,6 +20,10 @@ const optimizer = Ipopt.Optimizer(print_level=0, fixed_variable_treatment="make_
 const config = MOIT.TestConfig(atol=1e-4, rtol=1e-4,
                                optimal_status=MOI.LOCALLY_SOLVED)
 
+@testset "MOI API tests" begin
```
blegat: The test set name should be `SolverName` according to the testing guideline:
http://www.juliaopt.org/MathOptInterface.jl/dev/apimanual/#Testing-guideline-1

test/MOI_wrapper.jl
ccoffrin: Note that, indentation spacing is not consistent in the rest of this file.

test/MOI_wrapper.jl
ccoffrin: I stole this from Gurobi.jl, you might want to post an issue there, https://github.com/JuliaOpt/Gurobi.jl/blob/master/test/MOI_wrapper.jl#L82

PR #168, opened 2019-06-02
Switch to Project.toml and drop Julia 0.7

mlubin:


Project.toml
```diff
@@ -0,0 +1,23 @@
+name = "Ipopt"
+uuid = "b6b21f68-93f8-5de0-b562-5493be1d77c9"
```
blegat: I would also add `repo = "https://github.com/JuliaOpt/Ipopt.jl.git"` and `version = "..."` for consistency with other solvers

PR #170, opened 2019-06-06
Updates for MOI 0.9

mlubin:
Do not merge until MOI 0.9 is tagged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ sudo: false
 addons:
     apt_packages:
         - gfortran
+# TODO: Remove before merging.
+before_script:
+    - julia -e 'using Pkg; Pkg.add(PackageSpec(name="MathOptInterface", rev="master"))'
 after_success:
     - echo $TRAVIS_JULIA_VERSION
     - julia -e 'Pkg.add("Coverage"); cd(Pkg.dir("Ipopt")); using Coverage; Coveralls.submit(process_folder())'
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,6 @@ sudo: false
 addons:
     apt_packages:
         - gfortran
-# TODO: Remove before merging.
-before_script:
-    - julia -e 'using Pkg; Pkg.add(PackageSpec(name="MathOptInterface", rev="master"))'
 after_success:
     - echo $TRAVIS_JULIA_VERSION
     - julia -e 'Pkg.add("Coverage"); cd(Pkg.dir("Ipopt")); using Coverage; Coveralls.submit(process_folder())'
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,7 @@ MathProgBase = "fdba3010-5040-5b88-9595-932c9decdf73"
 
 [compat]
 BinaryProvider = "≥ 0.5.3"
-MathOptInterface = "~0.8.1"
+MathOptInterface = "0.9"
 MathProgBase = "~0.5.0, ~0.6, ~0.7"
 julia = "1"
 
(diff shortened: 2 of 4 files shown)


src/MOI_wrapper.jl
```diff
@@ -717,13 +781,12 @@ function MOI.optimize!(model::Optimizer)
                      for v in model.variable_info]
 
     for (name,value) in model.options
-        sname = string(name)
-        if match(r"(^resto_)", sname) != nothing
-            sname = replace(sname, r"(^resto_)", "resto.")
-        end
         addOption(model.inner, sname, value)
```
blegat: `sname` is not defined here

Project.toml
```diff
@@ -12,7 +12,7 @@ MathProgBase = "fdba3010-5040-5b88-9595-932c9decdf73"
 
 [compat]
 BinaryProvider = "≥ 0.5.3"
-MathOptInterface = "~0.8.1"
+MathOptInterface = "0.9"
```
ccoffrin: Should this be "~0.9"?

Wikunia:
According to https://github.com/JuliaOpt/MathOptInterface.jl/issues/736
`DualObjectiveValue` should be implemented which isn't the case as it's not supported.
Therefore the test config is changed. Is this the way it should be handled or should it return `nothing` or something like that? The requirements read like the function should exist to support MOI v0.9

blegat:
This was discussed on Gitter on June 03.
The guideline if it's not implemented by a solver is to use `MOIU.get_fallback`. In this case, to quote @mlubin in Gitter (June 03 13h31):
> The bigger issue is that dual objective is not a well-defined concept for NLP solvers in general, as far as I'm aware. We can compute it manually if all constraints are linear, but that's about it.

So we could implement it and call `MOIU.get_fallback` if the problem is linear the throw an error if it's not.

Wikunia:
Thanks @blegat !

mlubin:
I don't expect to have time to tidy this up and merge before the weekend, so it's up for grabs until then.

blegat:
@mlubin Good to merge and release as Ipopt v0.6.1 ?

.travis.yml
```diff
@@ -11,6 +11,9 @@ sudo: false
 addons:
     apt_packages:
         - gfortran
+# TODO: Remove before merging.
```
mlubin: This should be removed

Project.toml
mlubin: Yes

blegat: Both `"0.9"` and `"~0.9"` are equivalent, no ?: https://julialang.github.io/Pkg.jl/v1/compatibility/
For `"0.9"`, it's ok if you don't change the left-most nonzero so the interval is `[0.9, 0.10)`.
For `"~0.9"~, it's ok if you only modify patch so the interval is `[0.9, 0.10)` as well.

mlubin: > Caret specifiers are the default and hence 1.2.3 == ^1.2.3
> PkgB = "^1.2"   # [1.2.0, 2.0.0)

If I understand correctly, "0.9" would allow anything up to 1.0.

blegat: > A caret specifier allows upgrade that would be compatible according to semver. An updated dependency is considered compatible if the new version does not modify the left-most non zero digit in the version specifier.

For "1.2.3", the left-most is "1" so "1.3" **does** not modify the left-most nonzero.
For "0.9", the left-most is "9" so "0.10" **does** modify the left-most nonzero.

ccoffrin:
Thanks Miles!!!

PR #172, opened 2019-06-25
ConstraintDual for quadratic constraints

### LINKED ISSUE #171: Acessing constraint dual value
I have the same test optimization model in GAMS and Julia.

GAMS using Ipopt calculates dual values of constraints normally and in Julia I get error:
"ArgumentError: ModelLike of type Ipopt.Optimizer does not support accessing the attribute MathOptInterface.ConstraintDual(1)"

mlubin:
Fixes #171 

However, I'm hesitant to merge before we have an MOI convention for the sign of the duals and tests.
@schepy

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -907,41 +907,40 @@ function MOI.get(model::Optimizer, ::MOI.ConstraintPrimal,
     return model.inner.x[vi.value]
 end
 
-function MOI.get(model::Optimizer, ::MOI.ConstraintDual,
-                 ci::MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64},
-                                         MOI.LessThan{Float64}})
-    if model.inner === nothing
-        error("ConstraintDual not available.")
+macro define_constraint_dual(function_type, set_type, prefix)
+    constraint_array = Symbol(string(prefix) * "_constraints")
+    offset_function = Symbol(string(prefix) * "_offset")
+    quote
+        function MOI.get(model::Optimizer, ::MOI.ConstraintDual,
+                         ci::MOI.ConstraintIndex{$function_type, $set_type})
+            if model.inner === nothing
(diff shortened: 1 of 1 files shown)


src/MOI_wrapper.jl
```diff
@@ -907,41 +907,40 @@ function MOI.get(model::Optimizer, ::MOI.ConstraintPrimal,
...
-    if model.inner === nothing
-        error("ConstraintDual not available.")
+macro define_constraint_dual(function_type, set_type, prefix)
+    constraint_array = Symbol(string(prefix) * "_constraints")
+    offset_function = Symbol(string(prefix) * "_offset")
+    quote
+        function MOI.get(model::Optimizer, ::MOI.ConstraintDual,
+                         ci::MOI.ConstraintIndex{$function_type, $set_type})
+            if model.inner === nothing
+                error("ConstraintDual not available.")
```
blegat: Why no throw `MOI.OptimizeNotCalled()` ?

src/MOI_wrapper.jl
```diff
@@ -907,41 +907,40 @@ function MOI.get(model::Optimizer, ::MOI.ConstraintPrimal,
...
+            end
+            if !(1 <= ci.value <= length(model.$(constraint_array)))
+                error("Invalid constraint index ", ci.value)
```
blegat: Why not throw `MOI.InvalidIndex`, even better is call `MOI.throw_is_not_valid` to reuse `MOI.is_valid`

src/MOI_wrapper.jl
blegat: Moving to issue to https://github.com/JuliaOpt/Ipopt.jl/issues/174

src/MOI_wrapper.jl
blegat: Moving it to https://github.com/JuliaOpt/Ipopt.jl/issues/175

PR #176, opened 2019-07-15
 Implement dual warmstart

blegat:
Closes https://github.com/JuliaOpt/Ipopt.jl/issues/164
Requires https://github.com/JuliaOpt/Ipopt.jl/pull/172
cc @ccoffrin

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -4,13 +4,15 @@ const MOI = MathOptInterface
 mutable struct VariableInfo
     lower_bound::Float64  # May be -Inf even if has_lower_bound == true
     has_lower_bound::Bool # Implies lower_bound == Inf
+    lower_bound_start::Union{Nothing, Float64}
     upper_bound::Float64  # May be Inf even if has_upper_bound == true
     has_upper_bound::Bool # Implies upper_bound == Inf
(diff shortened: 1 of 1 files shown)


src/MOI_wrapper.jl
```diff
@@ -24,6 +26,13 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     quadratic_le_constraints::Vector{Tuple{MOI.ScalarQuadraticFunction{Float64}, MOI.LessThan{Float64}}}
     quadratic_ge_constraints::Vector{Tuple{MOI.ScalarQuadraticFunction{Float64}, MOI.GreaterThan{Float64}}}
     quadratic_eq_constraints::Vector{Tuple{MOI.ScalarQuadraticFunction{Float64}, MOI.EqualTo{Float64}}}
+    linear_le_dual_start::Vector{Union{Nothing, Float64}}
```
mlubin: To avoid parallel arrays and for consistency with variables, I'd prefer to structure this as
```julia
mutable struct ConstraintInfo{F, S}
    func::F
    set::S
    dual_start::Union{Nothing, Float64}
end
...
    linear_le_constraints::Vector{ConstraintInfo{MOI.ScalarAffineFunction{Float64}, MOI.LessThan{Float64}}
```

PR #182, opened 2019-10-14
Make tests pass with MOI 0.9.5

rschwarz:
I made some additions to the implementation (around `ResultCount`) but mostly excluded more MOI unit tests.

### DIFF
--- a/test/MOI_wrapper.jl
+++ b/test/MOI_wrapper.jl
@@ -59,6 +59,7 @@ end
                "number_threads", # NumberOfThreads not supported
                "delete_nonnegative_variables", # get ConstraintFunction n/a.
                "update_dimension_nonnegative_variables", # get ConstraintFunction n/a.
+               "delete_soc_variables", # VectorOfVar. in SOC not supported
                ]
     MOIT.unittest(bridged, config, exclude)
 end
--- a/test/MOI_wrapper.jl
+++ b/test/MOI_wrapper.jl
@@ -61,6 +61,8 @@ end
                "update_dimension_nonnegative_variables", # get ConstraintFunction n/a.
                "delete_soc_variables", # VectorOfVar. in SOC not supported
                "solve_result_index", # DualObjectiveValue not supported
+               "solve_single_variable_dual_min", # Var is fixed, Ipopt returns 0 dual?
+               "solve_single_variable_dual_max", # Var is fixed, Ipopt returns 0 dual?
                ]
     MOIT.unittest(bridged, config, exclude)
 end
--- a/test/MOI_wrapper.jl
+++ b/test/MOI_wrapper.jl
@@ -30,8 +30,9 @@ end
 end
 
 @testset "Unit" begin
-    bridged = MOIB.full_bridge_optimizer(Ipopt.Optimizer(print_level=0),
-                                         Float64)
+    bridged = MOIB.full_bridge_optimizer(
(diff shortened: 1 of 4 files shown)


rschwarz:
Two unit tests still fail because the values for `ConstraintDual` differ from the expectation. Not sure what is wrong with that.

test/MOI_wrapper.jl
```diff
@@ -55,7 +55,12 @@ end
                "solve_singlevariable_obj", # loadfromstring!
                "solve_objbound_edge_cases", # ObjectiveBound not supported.
                "solve_affine_deletion_edge_cases", # Deleting not supported.
-               "solve_unbounded_model" # `NORM_LIMIT`
+               "solve_unbounded_model", # `NORM_LIMIT`
+               "number_threads", # NumberOfThreads not supported
+               "delete_nonnegative_variables", # get ConstraintFunction n/a.
+               "update_dimension_nonnegative_variables", # get ConstraintFunction n/a.
+               "delete_soc_variables", # VectorOfVar. in SOC not supported
+               "solve_result_index", # DualObjectiveValue not supported
```
blegat: We could run it with `duals=false` in the config.

rschwarz: But that would deactivate even more (partial) tests overall, in the unit tests?

rschwarz: Or do you suggest running the tests twice, with different configs?

blegat: No forget about it, we really need to add a config option to disable just DualObjectiveValue tests

rschwarz: Or, (slightly off-topic), the tests itself could query the corresponding `supports` method and then skip part of the test.

blegat: We don't want that as a solver that forgot to define supports won't notice, we need the solver to explicitly declare that it does not support it in the tests

rschwarz:
> Two unit tests still fail because the values for `ConstraintDual` differ from the expectation. Not sure what is wrong with that.

I looked a little more into these tests. The model is minimizing/maximizing a single variable `1 <= x <= 1`. 

I'm guessing that Ipopt considers it fixed, and therefore does not provide proper duals (just 0)?
Not sure yet whether that happens in Ipopt or the Julia wrapper, or the MOI wrapper.

If I change the test in MOI so that (for minimization) the variable is bounded `1 <= x <= 2`, the dual value is correct and the test passes. So I wonder whether the two tests should be changed in MOI to something more agreeable?

rschwarz:
> I'm guessing that Ipopt considers it fixed, and therefore does not provide proper duals (just 0)?

See also the description of the parameter [fixed_variable_treatment](https://www.coin-or.org/Ipopt/documentation/node44.html#SECTION000114060000000000000) in the Ipopt docs:

> Determines how fixed variables should be handled.
The main difference between those options is that the starting point in the "make_constraint" case still has the fixed variables at their given values, whereas in the case "make_parameter" the functions are always evaluated with the fixed values for those variables. Also, for "relax_bounds", the fixing bound constraints are relaxed (according to" bound_relax_factor"). For both "make_constraints" and "relax_bounds", bound multipliers are computed for the fixed variables. The default value for this string option is "make_parameter".

So, with the default value `make_parameter`, no multipliers are computed.

rschwarz:
Now it should be good to go, without compromises.

PR #193, opened 2020-05-02
Use Ipopt_jll binaries on Julia pre v1.3 too

blegat:
@dpo What is the use of `Libdl.dlpath` in:
https://github.com/JuliaOpt/Ipopt.jl/blob/5efbc4e15ac2b565b6c7880ed404159a6289a890/src/Ipopt.jl#L11
`libipopt` is already an absolute path.


dpo:
Not sure it's necessary. I was mostly trying to preserve the variable names in Ipopt.jl.

dpo:
ps: there's an updated https://github.com/JuliaBinaryWrappers/Ipopt_jll.jl/releases

blegat:
> ps: there's an updated https://github.com/JuliaBinaryWrappers/Ipopt_jll.jl/releases

Indeed, that's the one I used to generate the `build.jl`.

deps/build.jl
```diff
@@ -1,4 +1,8 @@
-using BinaryProvider # requires BinaryProvider 0.3.0 or later
+if VERSION >= v"1.3" && !haskey(ENV, "JULIA_CLP_LIBRARY_PATH")
```
amontoison, line 1: `JULIA_IPOPT_LIBRARY_PATH` ?

blegat: Thanks, fixed

Issue #185, opened 2019-11-06
Project the starting point to variable bounds


mlubin:
The starting point should at least respect the variable bounds. Currently we're using 0.0 even if it's infeasible:

https://github.com/JuliaOpt/Ipopt.jl/blob/085b0177de3042c4a9b965e6f70cd61c2b5c22a5/src/MOI_wrapper.jl#L887-L889

Ref https://discourse.julialang.org/t/some-jump-variables-values-are-out-of-bounds/26008/6

Issue #183, opened 2019-10-29
Silent is ignored


mlubin:
The model correctly supports getting and setting `MOI.Silent`, but it's ignored at solve time.

PR #194, opened 2020-05-04
MOI.Silent suppresses output, project starting point into defined bounds

mewilhel:
Should address #183 and #185. 

- Causes MOI.Silent to suppress output by setting print_level to zero 0. This is overridden by a if print_level parameter is specifically set.
- Default starting point: For two-sided variable constraints, the starting point is set to the midpoint. For one-sided variable constraints, the starting point is set to the bound if 0.0 is infeasible.  For unconstrained variables, a starting value of 0.0 is used.
- Fix a typo in interpreting status.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -893,11 +893,11 @@ function MOI.optimize!(model::Optimizer)
             if !(v.has_lower_bound && v.has_upper_bound)
                 continue
             elseif v.has_lower_bound && v.has_upper_bound
-                model.inner.x[i] = 0.5*(v.lower_bound + x.upper_bound)
+                model.inner.x[i] = 0.5*(v.lower_bound + v.upper_bound)
             elseif v.has_lower_bound
-                model.inner.x[i] = max(0.0, x.lower_bound)
+                model.inner.x[i] = max(0.0, v.lower_bound)
             elseif v.has_upper_bound
-                model.inner.x[i] = min(0.0, x.upper_bound)
+                model.inner.x[i] = min(0.0, v.upper_bound)
             end
         end
     end
@@ -922,7 +922,7 @@ function MOI.optimize!(model::Optimizer)
     model.inner.mult_x_U = [v.upper_bound_dual_start === nothing ? 0.0 : v.lower_bound_dual_start
                             for v in model.variable_info]
 
-    model.silent && addOption(model.inner, "print_level ", 0)
+    model.silent && addOption(model.inner, "print_level", 0)
 
     for (name, value) in model.options
         addOption(model.inner, name, value)
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -885,17 +885,19 @@ function MOI.optimize!(model::Optimizer)
     end
 
     # If nothing is provided, the default starting value is 0.0.
-    if v.start !== nothing
-        model.inner.x = v.start
-    else
-        model.inner.x = fill(0.0, num_variables)
-        for (i, v) in enumerate(model.variable_info)
-            if v.has_lower_bound && v.has_upper_bound
+    model.inner.x = fill(0.0, num_variables)
+    for (i, v) in enumerate(model.variable_info)
+        if v.start !== nothing
+            model.inner.x[i] = v.start
+        else
+            if !(v.has_lower_bound && v.has_upper_bound)
+                continue
+            elseif v.has_lower_bound && v.has_upper_bound
                 model.inner.x[i] = 0.5*(v.lower_bound + x.upper_bound)
             elseif v.has_lower_bound
-                model.inner.x[i] = x.lower_bound
+                model.inner.x[i] = max(0.0, x.lower_bound)
             elseif v.has_upper_bound
-                model.inner.x[i] = x.upper_bound]
+                model.inner.x[i] = min(0.0, x.upper_bound)
             end
         end
     end
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -885,8 +885,20 @@ function MOI.optimize!(model::Optimizer)
     end
 
     # If nothing is provided, the default starting value is 0.0.
-    model.inner.x = [v.start === nothing ? 0.0 : v.start
-                     for v in model.variable_info]
+    if v.start !== nothing
+        model.inner.x = v.start
+    else
+        model.inner.x = fill(0.0, num_variables)
+        for (i, v) in enumerate(model.variable_info)
+            if v.has_lower_bound && v.has_upper_bound
+                model.inner.x[i] = 0.5*(v.lower_bound + x.upper_bound)
+            elseif v.has_lower_bound
+                model.inner.x[i] = x.lower_bound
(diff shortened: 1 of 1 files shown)


src/MOI_wrapper.jl
```diff
@@ -885,8 +885,22 @@ function MOI.optimize!(model::Optimizer)
     end
 
     # If nothing is provided, the default starting value is 0.0.
-    model.inner.x = [v.start === nothing ? 0.0 : v.start
-                     for v in model.variable_info]
+    model.inner.x = fill(0.0, num_variables)
```
blegat, line 888: Why not `zeros(num_variables)` ?

src/MOI_wrapper.jl
```diff
@@ -885,8 +885,22 @@ function MOI.optimize!(model::Optimizer)
...
+    for (i, v) in enumerate(model.variable_info)
+        if v.start !== nothing
+            model.inner.x[i] = v.start
+        else
+            if !(v.has_lower_bound && v.has_upper_bound)
```
blegat, line 893: Shouldn't it be `!v.has_lower_bound && !v.has_upper_bound` ? Otherwise, the last two `elseif` can't be reached. It seems we can just drop this `continue` as this is just the complement case of the other three cases.

src/MOI_wrapper.jl
```diff
@@ -885,8 +885,22 @@ function MOI.optimize!(model::Optimizer)
...
+                continue
+            elseif v.has_lower_bound && v.has_upper_bound
+                model.inner.x[i] = 0.5*(v.lower_bound + v.upper_bound)
```
blegat, line 896: For consistency with the other two cases, we can keep `0.0` if it is in the interval ?

src/MOI_wrapper.jl
mewilhel: Fair point.

src/MOI_wrapper.jl
mewilhel: Fixed.

src/MOI_wrapper.jl
mewilhel: Works for me, included in update.

src/MOI_wrapper.jl
```diff
@@ -885,8 +885,27 @@ function MOI.optimize!(model::Optimizer)
...
 
     # If nothing is provided, the default starting value is 0.0.
-    model.inner.x = [v.start === nothing ? 0.0 : v.start
-                     for v in model.variable_info]
+    model.inner.x = zeros(num_variables)
+    for (i, v) in enumerate(model.variable_info)
+        if v.start !== nothing
+            model.inner.x[i] = v.start
+        else
+            if !v.has_lower_bound && !v.has_upper_bound
```
blegat, line 893: I would find it much more readable to have
```julia
            if v.has_lower_bound && v.has_upper_bound
                # mid(lower_bound, upper_bound, 0.0), if lower_bound <= upper_bound
                if (0.0 <= v.lower_bound)
                    model.inner.x[i] = v.lower_bound
                elseif (v.upper_bound <= 0.0)
                    model.inner.x[i] = v.upper_bound
                end
            elseif v.has_lower_bound
                model.inner.x[i] = max(0.0, v.lower_bound)
            elseif v.has_upper_bound
                model.inner.x[i] = min(0.0, v.upper_bound)
            end
```

src/MOI_wrapper.jl
```diff
@@ -885,8 +885,27 @@ function MOI.optimize!(model::Optimizer)
...
+                continue
+            elseif v.has_lower_bound && v.has_upper_bound
+                # mid(lower_bound, upper_bound, 0.0), if lower_bound <= upper_bound
+                if (0.0 <= v.lower_bound)
```
blegat, line 897: You don't need parentheses around the condition

src/MOI_wrapper.jl
```diff
@@ -885,8 +885,27 @@ function MOI.optimize!(model::Optimizer)
...
-                     for v in model.variable_info]
+    model.inner.x = zeros(num_variables)
+    for (i, v) in enumerate(model.variable_info)
+        if v.start !== nothing
+            model.inner.x[i] = v.start
+        else
+            if !v.has_lower_bound && !v.has_upper_bound
+                continue
+            elseif v.has_lower_bound && v.has_upper_bound
+                # mid(lower_bound, upper_bound, 0.0), if lower_bound <= upper_bound
```
blegat, line 896: Isn't this comment outdated ? What do you refer to by `mid` ?

mewilhel: I'll delete it. I was referring to the middle value of lower_bound, upper_bound, and 0.0

src/MOI_wrapper.jl
mewilhel: Done.

src/MOI_wrapper.jl
mewilhel: Done.

src/MOI_wrapper.jl
```diff
@@ -885,8 +885,24 @@ function MOI.optimize!(model::Optimizer)
     end
 
     # If nothing is provided, the default starting value is 0.0.
-    model.inner.x = [v.start === nothing ? 0.0 : v.start
-                     for v in model.variable_info]
+    model.inner.x = zeros(num_variables)
```
odow, line 888: Given how common it is for people to write `1 / x` and then complain. Do we want to start at `0`? What about starting at `1`? That also avoids issues with `log`.

mlubin: The value `1` has no meaning if we don't know anything about the scale of the variables. Projecting 0 onto the bounds of the variables is consistent with what we did in JuMP 0.18 and with AMPL I believe. To address the complaints we could have better error messages telling users to set their own starting values.

blegat: Then, they will have issues with `1 / (x - 1)` and `log(x - 1)`. Using `(lb + ub) / 2` for interval, `lb + 1` for lower bound only and `ub - 1` for upper bound only might fix most issues ? If the user uses `1 / (x - a)`, the bounds on the variable might be `x <= a` or `x >= a`.

mewilhel: A fair amount of common domain violations and differentiability issues (e.g. sqrt, 1/x, log, x^-3) occur at zero. So maybe we could do the following, to keep all the initial guesses nonzero.

- use `ub/2` for both the interval case 
- use `ub/2` for the upper bound only case 
- use `lb/2` for the lower bound only case

mlubin: I think trying to be clever here will make it harder for people to debug these issues. For example, suppose someone has a working model, removes a bound on a variable, and then gets solver crashes because of the new starting point.

mewilhel: Fair point. I believe the current behavior is to throw either a MOI.INVALID_MODEL, MOI.NUMERICAL_ERROR, or some limit termination status code when a domain violation is encountered. All of which seems pretty transparent.

odow:
@mewilhel is this ready to merge?

mewilhel:
@odow Yeah, this should be ready to go.

PR #200, opened 2020-05-29
fixes #199

### LINKED ISSUE #199: Accessing MOI.ObjectiveFunctionType
Currently Ipopt.jl doesn't supports accessing certain MOI model attributes such as `ObjectiveFunctionType` (read a similar issue https://github.com/JuliaOpt/Ipopt.jl/issues/171)

AKS1996:


AKS1996:
still a long way to fix #199

src/MOI_wrapper.jl
```diff
@@ -132,13 +132,63 @@ end
...
+            elseif info.lower_bound == info.upper_bound
+                push!(constraints, (MOI.SingleVariable, MOI.EqualTo{Float64}))
+            else
+                push!(constraints, (MOI.SingleVariable, MOI.Interval{Float64}))
+            end
+        end 
+    end
+
+    # handling model constraints separately
+    if size(model.linear_le_constraints)[1] > 0
```
blegat, line 169: Just use `!isempty(model.linear_le_constraints)`

src/MOI_wrapper.jl
```diff
@@ -132,13 +132,63 @@ end
...
 
 function MOI.get(model::Optimizer, ::MOI.ListOfVariableIndices)
     return [MOI.VariableIndex(i) for i in 1:length(model.variable_info)]
 end
 
 
+function MOI.get(model::Optimizer, ::MOI.ListOfConstraints)
+    constraints = Set{Tuple{DataType, DataType}}()
+    for info in model.variable_info
+        if !info.has_lower_bound && !info.has_upper_bound
```
blegat, line 147: You need to return the constraint types that were added with `add_constraint`. Independently on the value of the bounds.
You can just do something like
```julia
if info.has_lower_bound
    push!(constraints, (MOI.SingleVariable, MOI.LessThan{Float64}))
end
if info.has_upper_bound
    push!(constraints, (MOI.SingleVariable, MOI.GreaterThan{Float64}))
end
if info.fixed
    push!(constraints, (MOI.SingleVariable, MOI.EqualTo{Float64}))
end
```

PR #204, opened 2020-06-13
minor change in view of #199

AKS1996:
struct `VariableInfo` has `is_fixed` variable. confused it earlier with `fixed`

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -136,6 +136,13 @@ MOI.get(model::Optimizer, ::MOI.ObjectiveFunctionType) = typeof(model.objective)
 
 MOI.get(model::Optimizer, ::MOI.NumberOfVariables) = length(model.variable_info)
 
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64}, MOI.LessThan{Float64}}) = length(model.linear_le_constraints)
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64}, MOI.EqualTo{Float64}}) = length(model.linear_eq_constraints)
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64}, MOI.GreaterThan{Float64}}) = length(model.linear_ge_constraints)
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.SingleVariable, MOI.LessThan{Float64}}) = length(filter(e -> e.has_upper_bound, model.variable_info))
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.SingleVariable, MOI.EqualTo{Float64}}) = length(filter(e -> e.is_fixed, model.variable_info))
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.SingleVariable, MOI.GreaterThan{Float64}}) = length(filter(e -> e.has_lower_bound, model.variable_info))
+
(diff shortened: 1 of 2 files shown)


AKS1996:
added a few more attributes, as I wasn't able to get all of them using 
```
MOI.instantiate(Ipopt.Optimizer, with_bridge_type=Float64)
```

src/MOI_wrapper.jl
```diff
@@ -179,6 +179,56 @@ function MOI.get(model::Optimizer, ::MOI.ListOfConstraints)
...
+    end
+    return indices
+end
+
+
+function MOI.get(
+    model::Ipopt.Optimizer,
+    ::MOI.ListOfConstraintIndices{MOI.ScalarAffineFunction{Float64}, MOI.EqualTo{Float64}}
+) 
+    indices = MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64}, MOI.EqualTo{Float64}}[]
```
blegat, line 199: You can do `MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64}, MOI.EqualTo{Float64}}.(eachindex(model.linear_eq_constraints))`

blegat:
Please add a test as suggested in https://github.com/jump-dev/Ipopt.jl/pull/200#pullrequestreview-421292620

src/MOI_wrapper.jl
AKS1996: added

AKS1996:
trying to add but it [breaks at this test](https://github.com/jump-dev/MathOptInterface.jl/blob/master/src/Test/modellike.jl#L496). will it require adding methods to support vector constraints in Ipopt?

blegat:
> will it require adding methods to support vector constraints in Ipopt?

No just use `MOIB.instantiate(Ipopt.Optimizer, with_bridge_type=Float64)` instead of `optimizer`

AKS1996:
works now!

src/MOI_wrapper.jl
```diff
@@ -136,6 +136,13 @@ MOI.get(model::Optimizer, ::MOI.ObjectiveFunctionType) = typeof(model.objective)
 
 MOI.get(model::Optimizer, ::MOI.NumberOfVariables) = length(model.variable_info)
 
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64}, MOI.LessThan{Float64}}) = length(model.linear_le_constraints)
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64}, MOI.EqualTo{Float64}}) = length(model.linear_eq_constraints)
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.ScalarAffineFunction{Float64}, MOI.GreaterThan{Float64}}) = length(model.linear_ge_constraints)
+MOI.get(model::Optimizer, ::MOI.NumberOfConstraints{MOI.SingleVariable, MOI.LessThan{Float64}}) = length(filter(e -> e.has_upper_bound, model.variable_info))
```
blegat, line 142: Use `count` instead

test/MOI_wrapper.jl
```diff
@@ -95,3 +95,7 @@ MOI.empty!(optimizer)
 @testset "MOI NLP tests" begin
     MOIT.nlptest(optimizer, config)
 end
+
+@testset "Testing attributes" begin
```
blegat, line 99: `Testing getters` would better describe it imo

AKS1996: changed

src/MOI_wrapper.jl
AKS1996: changed to count

PR #214, opened 2020-07-27
Implement MOI.is_valid

blegat:
Closes https://github.com/jump-dev/Ipopt.jl/issues/211


src/MOI_wrapper.jl
```diff
@@ -411,12 +411,18 @@ end
 function MOI.add_variables(model::Optimizer, n::Int)
     return [MOI.add_variable(model) for i in 1:n]
 end
-
-function check_inbounds(model::Optimizer, vi::MOI.VariableIndex)
```
odow, line 415: Seems like you missed a case of where this was being called.

PR #225, opened 2020-10-27
Use obj_scaling_factor to set objective sense

### LINKED ISSUE #184: investigate obj_scaling_factor for maximizing
The [`obj_scaling_factor`](https://www.coin-or.org/Ipopt/documentation/node43.html#SECTION000113010000000000000) parameter looks like it might be a better solution to handle maximizing in the MOI wrapper than what we do currently because Ipopt's output will display the original objective and not a negated objective.

odow:
Closes #184 

Avoids having to multiply objective function/gradient by -1 if `MAX_SENSE`, and makes it so the Ipopt log now displays the correct objective value, minimizing confusion for users.

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -694,6 +694,11 @@ end
 
 # TODO(odow): this definition is incorrect. Should probably be
 # MOI.supports(::Optimizer, ::MOI.ConstrainDualStart, ::Type{<:MOI.ConstraintIndex})
+_dual_start(::Optimizer, ::Nothing, ::Int) = nothing
+function _dual_start(model::Optimizer, value::Real, scale::Int = 1)
+    return _dual_multiplier(model) * value * scale
+end
+
 function MOI.supports(
     ::Optimizer,
     ::MOI.ConstraintDualStart,
@@ -710,7 +715,7 @@ function MOI.set(
     value::Union{Real, Nothing},
 )
     MOI.throw_if_not_valid(model, ci)
-    model.variable_info[ci.value].lower_bound_dual_start = value
+    model.variable_info[ci.value].lower_bound_dual_start = _dual_start(model, value)
     return
 end
 
@@ -732,7 +737,7 @@ function MOI.set(
     value::Union{Real, Nothing},
 )
     MOI.throw_if_not_valid(model, ci)
-    model.variable_info[ci.value].upper_bound_dual_start = value
+    model.variable_info[ci.value].upper_bound_dual_start = _dual_start(model, value)
     return
 end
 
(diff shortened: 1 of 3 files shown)


src/MOI_wrapper.jl
```diff
@@ -1356,32 +1366,72 @@ end
...
+function MOI.get(
+    model::Optimizer,
+    attr::MOI.ConstraintDual,
+    ci::MOI.ConstraintIndex{MOI.SingleVariable, MOI.LessThan{Float64}},
+)
     MOI.check_result_index_bounds(model, attr)
     MOI.throw_if_not_valid(model, ci)
-    # MOI convention is for feasible LessThan duals to be nonpositive.
-    return -1 * model.inner.mult_x_U[ci.value]
+    # Due to differences between Ipopt's and MOI's definition of duality, it
```
mlubin, line 1376: Why doesn't a similar approach with `_dual_multiplier` work here?

src/MOI_wrapper.jl
```diff
@@ -1326,19 +1332,23 @@ function MOI.get(model::Optimizer, attr::MOI.ConstraintPrimal,
...
+        )
             MOI.check_result_index_bounds(model, attr)
             if !(1 <= ci.value <= length(model.$(constraint_array)))
                 error("Invalid constraint index ", ci.value)
             end
-            # TODO: Unable to find documentation in Ipopt about the signs of duals.
-            # Rescaling by -1 here seems to pass the MOI tests.
-            return -1 * model.inner.mult_g[ci.value + $offset_function(model)]
+            s = -_dual_multiplier(model)
+            return s * model.inner.mult_g[ci.value + $offset_function(model)]
```
mlubin, line 1351: The same transformations need to apply for the dual starts.

src/MOI_wrapper.jl
odow: I couldn't get all the tests passing using `_dual_multiplier`. Digging deeper, maybe this could be regarded as a bug in Ipopt. It's like they handle `fixed_variable_treatment` before applying the `obj_scaling_factor`:
```Julia
using JuMP, Ipopt
set_optimizer_attribute(model, "fixed_variable_treatment", "make_constraint")
@variable(model, 1 <= x <= 1)
@objective(model, Max, x)
optimize!(model)
ipopt = backend(model).optimizer.model
ipopt.inner.mult_x_L  # [1.0]
ipopt.inner.mult_x_U  # [0.0]
```
But you should expect that `mult_x_U` is the binding constraint. Compare with:
```Julia
using JuMP, Ipopt
@variable(model, x)
@objective(model, Max, x)
@constraint(model, cu, x <= 1)
@constraint(model, cl, x >= 1)
optimize!(model)
ipopt = backend(model).optimizer.model
ipopt.inner.mult_g  # [-1.0, 0.0]
```

src/MOI_wrapper.jl
odow: ConstraintDualStart's aren't even tested. In fact, most of the interface isn't tested.

src/MOI_wrapper.jl
odow:
```Julia
julia> using JuMP, Ipopt

julia> function fixed_variable_bug(treatment = "make_constraint")
           set_silent(model)
           set_optimizer_attribute(model, "fixed_variable_treatment", treatment)
           @variable(model, 1 <= x <= 1)

           @objective(model, Max, x)
           optimize!(model)
           ipopt = backend(model).optimizer.model.inner
           max_solution = (ipopt.mult_x_L[1], ipopt.mult_x_U[1])

           @objective(model, Min, x)
           optimize!(model)
           ipopt = backend(model).optimizer.model.inner
           min_solution = (ipopt.mult_x_L[1], ipopt.mult_x_U[1])

           return max_solution, min_solution
       end
fixed_variable_bug (generic function with 2 methods)

julia> fixed_variable_bug("make_constraint")
((1.0, 0.0), (1.0, 0.0))

julia> fixed_variable_bug("relax_bounds")
((-6.303429437358755e-16, -1.0000000000000007), (1.085946963838724, 0.085946963838724))
```

blegat (approved):

PR #230, opened 2020-11-06
Fix error in add_constraint NaN handling for EqualTo

freemin7:
This looks like a copy paste error which was never caught in testing


mlubin:
Thanks!

freemin7:
No problem, i am reading the Ipopt code since i am taking inspiration to wrap another NLP solver.
During that i noticed you have not getters for ``ScalarQuadraticFunction`` constraints. I would expect to see type signatures like:
```
MOI.get(
    model::Optimizer,
    ::MOI.ListOfConstraintIndices{MOI.ScalarQuadraticFunction{Float64}, MOI.Interval{Float64}}
)
```

Is that on purpose?

mlubin:
No, probably not on purpose. The quadratic functions were added after the affine functions.

PR #232, opened 2020-11-09
Refactor Ipopt tests

odow:
First step towards cleaning up the MOI wrapper.


test/MOI_wrapper.jl
```diff
@@ -1,110 +1,144 @@
...
-               "number_threads", # NumberOfThreads not supported
-               "delete_nonnegative_variables", # get ConstraintFunction n/a.
-               "update_dimension_nonnegative_variables", # get ConstraintFunction n/a.
-               "delete_soc_variables", # VectorOfVar. in SOC not supported
-               "solve_result_index", # DualObjectiveValue not supported
-               ]
-    MOIT.unittest(bridged, config, exclude)
+function test_basicconstraint()
+    # TODO(odow): these basic constraint tests are broken.
+    # MOI.Test.basic_constraint_tests(OPTIMIZER, CONFIG)
```
blegat, line 53: What is broken?

odow: `Ipopt.Optimizer` doesn't support constraint names, querying functions, etc. The other tests pass because they use a `CachingOptimizer`.

Issue #247, opened 2020-11-23
Linking Ipopt with linear solvers


odow:
There are a lot of issues relating to linear solvers with Ipopt. #236 was a start at providing some documentation on how to use Ipopt with different solvers: https://github.com/jump-dev/Ipopt.jl#custom-installation. However, most are untested and I don't have a Windows machine. I've opened this issue to track which solver-platform combinations we have instructions for.

If you have clarifications for existing instructions, or if you have instructions for a missing solver-platform pair, please open a pull request.

Here are solver/platform combinations I have tested and got working so far:

* Pardiso Project
  - [ ] Windows
  - [x] Mac
  - [x] Linux  (Related issues: #106, #177)
* MKL Pardiso
  - [ ] Windows
  - [ ] Mac
  - [ ] Linux (Related issues: #195)
* HSL for Ipopt (MA27)
  - [ ] Windows (Related issues: #229)
  - [x] Mac
  - [x] Linux
* HSL (MA86, MA97)
  - [ ] Windows (Related issues: #192)
  - [x] Mac (Related issues: #197, #213)
  - [x] Linux

metab0t:
I try it on Windows
When I put `libhsl.dll` in a directory in my PATH and use Ipopt bundled with `Ipopt_jll`, following error is encountered.
```
Exception of type: OPTION_INVALID in file "IpAlgBuilder.cpp" at line 268:
 Exception message: Selected linear solver MA27 not available.
Tried to obtain MA27 from shared library "libhsl.dll", but the following error occured:
loadLib error: Do not know how to handle shared libraries on this operating system

EXIT: Invalid option encountered.
```
I suspect that dynamic loading of linear solver is unavailable for `Ipopt_jll` on Windows.

ksun46:
I use Mac 11.0.1. I have a customized Ipopt installed in `~/CoinOR/Ipopt`, and all third party lib files can be found in `~/CoinOR/Ipopt/build/lib` (including `libhsl.dylib`). 

I followed the custom installation section by setting the lib path to `~/CoinOR/Ipopt/build/lib` and executable path to `~/CoinOR/Ipopt/build/bin`, build, `using Ipopt`, and then call Ipopt with MA27 in JuMP. But It didn't work. I still saw
```
Exception of type: OPTION_INVALID in file "IpAlgBuilder.cpp" at line 268:
 Exception message: Selected linear solver MA27 not available.
[...]

PR #248, opened 2020-11-23
[README] Clarify custom installation instructions

odow:
@ksun46 did you make sure these environment variables were set before calling `using Ipopt`?

If you previously ran `using Ipopt`, you will need to open a new Julia session for the changes to take effect.

x-ref: https://github.com/jump-dev/Ipopt.jl/issues/247#issuecomment-731947083

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,8 +88,16 @@ import Pkg
 Pkg.build("Ipopt")
 ```
 
-**Very important note: before calling `using Ipopt` in any Julia session, you
-must set these environment variables.**
+**Very important note: you must set these environment variables before 
+calling `using Ipopt` in every Julia session.**
+
+For example:
+```julia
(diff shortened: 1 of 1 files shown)


ksun46:
@odow Yes I think so. I saw the **very important note** in the readme. 

Just to confirm, we do not need a customized Ipopt to use HSL linear solvers. For example, every time when I start a new project, I will activate the project, and I would like to add the latest Ipopt and other packages into the toml files by doing `Pkg.add("Ipopt")`. I DONT need to set environment variables (to point to the customized Ipopt) for every project and `Pkg.build("Ipopt")`, right? I just need to make sure `libhsl.dylib` can be found somewhere in the load path (even before I start Julia)? Thanks.

odow:
> Just to confirm, we do not need a customized Ipopt to use HSL linear solvers

Correct.

PR #252, opened 2020-11-24
Allow empty models to be created

### LINKED ISSUE #238: Ipopt fails to load an empty model
```Julia
model = Ipopt.Optimizer()
MOI.optimize!(model)

julia> MOI.optimize!(model)
ERROR: IPOPT: Failed to construct problem.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] createProblem(::Int64, ::Array{Float64,1}, ::Array{Float64,1}, ::Int64, ::Array{Float64,1}, ::Array{Float64,1}, ::Int64, ::Int64, ::Ipopt.var"#eval_f_cb#46"{Ipopt.Optimizer}, ::Ipopt.var"#eval_g_cb#48"{Ipopt.Optimizer}, ::Ipopt.var"#eval_grad_f_cb#47"{Ipopt.Optimizer}, ::Ipopt.var"#eval_jac_g_cb#49"{Ipopt.Optimizer,Array{Tuple{Int64,Int64},1}}, ::Ipopt.var"#eval_h_cb#50"{Ipopt.Optimizer,Array{Tuple{Int64,Int64},1}}) at /home/parallels/.julia/packages/Ipopt/bYzBL/src/Ipopt.jl:250
 [3] optimize!(::Ipopt.Optimizer) at /home/parallels/.julia/packages/Ipopt/bYzBL/src/MOI_wrapper.jl:1104
 [4] top-level scope at REPL[3]:1
```

odow:
Closes #238

### DIFF
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -321,13 +321,6 @@ function createProblem(
 )
     @assert n == length(x_L) == length(x_U)
     @assert m == length(g_L) == length(g_U)
-    if n == 0
-        # Ipopt cannot create a problem if there are no variables. As a hack,
-        # add a variable fixed to 0.
-        n = 1
-        push!(x_L, 0.0)
-        push!(x_U, 0.0)
-    end
     # Wrap callbacks
     eval_f_cb = @cfunction(
         eval_f_wrapper,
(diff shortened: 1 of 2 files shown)


frapac:
I am not sure to understand why this is useful. If the user does not provide any variable, I would expect the wrapper to return an error instead of adding silently a dummy variable.

mlubin:
I'd vote for catching this and reporting a more useful error than Ipopt's internal error.

odow:
@frapac is this better?

frapac:
Looks good to me! Thanks for adding this error, I think it's better to state things explicitly.

blegat (approved):

Issue #269, opened 2021-05-04
Typo in README: Linear Solvers


loicvh:
The Linux explanation for the installation of Pardiso contains a minor typo at step 4:

> Alternatively, if the library is located at /full/path/libpardiso.dylib

`.dylib` is the extension for MacOS, it should read

> Alternatively, if the library is located at /full/path/libpardiso.so

Also, it could be nice to explain how to add to the load path, e.g.,

```
export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/full/path/"
```
I can create a quick PR for this (really minor) issue.

odow:
Closed by #270 

Thanks @loicvh

PR #270, opened 2021-05-04
Fix typo in Linear solvers

loicvh:
Fix typo in #269 and show how to extend the load path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,8 +167,12 @@ _Tested on a clean install of Ubuntu 20.04._
    ```
 2. Download Pardiso from [https://www.pardiso-project.org](https://www.pardiso-project.org)
 3. Rename the file `libpardiso-XXXXX.so` to `libpardiso.so`
-4. Place the `libpardiso.so` library somewhere on your load path.
-   - Alternatively, if the library is located at `/full/path/libpardiso.dylib`,
+4. Place the `libpardiso.so` library somewhere on your load path, e.g., by adding to your
+   `.bashrc`:
+   ```
+   export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/full/path/"
+   ```
+   - Alternatively, if the library is located at `/full/path/libpardiso.so`,
      start Julia with `export LD_LIBRARY_PATH=/full/path; julia`
 5. Set the option `linear_solver` to `pardiso`:
    ```julia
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -167,8 +167,12 @@ _Tested on a clean install of Ubuntu 20.04._
    ```
 2. Download Pardiso from [https://www.pardiso-project.org](https://www.pardiso-project.org)
 3. Rename the file `libpardiso-XXXXX.so` to `libpardiso.so`
-4. Place the `libpardiso.so` library somewhere on your load path.
-   - Alternatively, if the library is located at `/full/path/libpardiso.dylib`,
+4. Place the `libpardiso.so` library somewhere on your load path, e.g., by adding to your
```
odow, line 170: I meant somewhere like `/opt`.

README.md
```diff
@@ -167,8 +167,12 @@ _Tested on a clean install of Ubuntu 20.04._
...
+   `.bashrc`:
+   ```
+   export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/full/path/"
+   ```
+   - Alternatively, if the library is located at `/full/path/libpardiso.so`,
      start Julia with `export LD_LIBRARY_PATH=/full/path; julia`
```
odow, line 176: To make this permanent, modify your `.bashrc` to include:
```
export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/full/path/"
```

odow (commented):
Thanks for the suggestions. Were they helpful in getting things working?

README.md
loicvh: > Thanks for the suggestions. Were they helpful in getting things working?

I know that it can be confusing how and why adding folders to the path, and since you mention the trick of appending `export LD_LIBRARY_PATH=/full/path`, I believe it can make life easier to the user.

You could also directly suggest the user to move `libpardiso.so` to `/opt`.

README.md
loicvh: > To make this permanent, modify your `.bashrc` to include:
> 
> ```
> export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/full/path/"
> ```

I made a modification following your suggestion in https://github.com/jump-dev/Ipopt.jl/pull/270/commits/708c47db37a9c12d894b45549b4070ccc25592d0

PR #274, opened 2021-06-13
[breaking] remove ability to access binary from Ipopt.jl

### LINKED ISSUE #261: Deprecate and remove __init__
This should be removed at some point.

https://github.com/jump-dev/Ipopt.jl/blob/31a7ac15675f73ff68ea02f07a0b9eebdd8f6aac/src/Ipopt.jl#L64-L91

AmplNLWriter doesn't need it: https://github.com/jump-dev/AmplNLWriter.jl/issues/77

odow:
This PR removes `amplexefun` and related things. In particular, the `__init__` method broke other packages by modifying `LD_LIBRARY_PATH`. Instead of providing a nice deprecation, it just throws an error. 

The original motivation for this was to provide a nice way to call the Ipopt binary for things like AmplNLWriter. Now that we have Ipopt_jll, Ipopt.jl no longer needs to do this job.

One exception is on Julia 1.0, but then we can just tell people to use `Ipopt.jl@0.6`.

Closes #261

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Ipopt"
 uuid = "b6b21f68-93f8-5de0-b562-5493be1d77c9"
 repo = "https://github.com/jump-dev/Ipopt.jl.git"
-version = "0.6.5"
+version = "0.7.0"
 
 [deps]
 BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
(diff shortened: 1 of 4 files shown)


blegat (approved):

odow:
Merging and tagging a new release because this started causing problems for the JuMP documentation

PR #275, opened 2021-06-22
Convert objective if it's not of the requested type

blegat:
It caused this issue https://github.com/jump-dev/DiffOpt.jl/issues/117

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -563,7 +563,7 @@ function MOI.get(
 end
 
 function MOI.get(model::Optimizer, ::MOI.ObjectiveFunction{T})::T where {T}
-    return model.objective
+    return convert(T, model.objective)
 end
 
 function MOI.set(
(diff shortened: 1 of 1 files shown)


src/MOI_wrapper.jl
```diff
@@ -562,7 +562,7 @@ function MOI.get(
     return MOI.GreaterThan{Float64}(model.variable_info[c.value].lower_bound)
 end
 
-function MOI.get(model::Optimizer, ::MOI.ObjectiveFunction)
+function MOI.get(model::Optimizer, ::MOI.ObjectiveFunction{T})::T where {T}
     return model.objective
```
odow, line 566:
```suggestion
    return convert(T, model.objective)
```

The annotation alone is pretty sneaky (I missed in the first read).

blegat: I just copy-pasted the definition from `MOI/src/Utilities/model.jl`.

odow: Then we should probably also change `model.jl` haha.

blegat: Agreed, this seems to aspect of the definition seems to be overlooked by many users so making it more visible is a good thing

PR #277, opened 2021-07-22
Add option Maximum_WallTime_Exceeded

qlete:
Add the new Ipopt option Maximum_WallTime_Exceeded, available since version 3.14.0. See [here](https://github.com/coin-or/Ipopt/blob/e4f829e7263cd05e55eafc3077edd05871f1ed21/src/Interfaces/IpReturnCodes_inc.h#L27).


odow:
Thanks! However, we currently ship 3.13 by default, so I would prefer not to expose the option until we update to 3.14. I opened an issue: https://github.com/jump-dev/Ipopt.jl/issues/278

odow:
Closing in favor of #278. I will do this when I update 3.14.4.

PR #292, opened 2021-11-09
Add MOI.SolverVersion

odow:


blegat:
I'd rather not implement it and exclude the test like with SCS. We should be consistent with that.

odow:
Yeah I did this before SCS.

PR #304, opened 2022-02-21
Fix edge cases for RawStatusString

blegat:
This would produce errors, e.g., when using `JuMP.solution_summary`.

### DIFF
--- a/test/MOI_wrapper.jl
+++ b/test/MOI_wrapper.jl
@@ -184,7 +184,7 @@ end
 
 function test_empty_optimize()
     model = Ipopt.Optimizer()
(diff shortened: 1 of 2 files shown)


test/MOI_wrapper.jl
```diff
@@ -184,10 +184,12 @@ end
 
 function test_empty_optimize()
     model = Ipopt.Optimizer()
+    @test MOI.get(model, MOI.RawStatusString()) == "The model has no variable"
```
odow, line 187:
```suggestion
    @test MOI.get(model, MOI.RawStatusString()) == "Optimize not called"
```

PR #307, opened 2022-02-22
[README] add more dependencies to linux HSL build

### LINKED ISSUE #300: MA86 compile or library failure on WSL/Ubuntu within Windows: undefined symbol: cgemv_
I'm trying to use MA86 in IPOPT on my Windows system, so I've resorted to compiling it in WSL (Ubuntu). Following the README instructions sort of worked (see further below), but when solving a model using IPOPT in Julia I get this error:

```
Exception of type: DYNAMIC_LIBRARY_FAILURE in file "Common/IpLibraryLoader.cpp" at line 67:
 Exception message: /home/niclas/julia-1.7.1/bin/libhsl.so: undefined symbol: cgemv_

EXIT: Library loading failure.
```

Possibly related: during the compilation I had the following problem during `./configure`:

```
configure: WARNING: No BLAS library found. make check will not work.
checking for METIS library... checking for metis_nodend_ in -L -lmetis... no
checking for metis_nodend_ in -lmetis... no
./configure: line 20798: hsl_metis_ok: command not found
configure: WARNING: No MeTiS library found. make check will not work.
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: creating src/Makefile
config.status: creating example/Fortran/Makefile
config.status: creating example/C/Makefile
config.status: executing depfiles commands
config.status: error: in `/home/niclas/hsl_ma86-1.6.0':
config.status: error: Something went wrong bootstrapping makefile fragments
    for automatic dependency tracking.  Try re-running configure with the
    '--disable-dependency-tracking' option to at least be able to build
    the package (albeit without support for automatic dependency tracking).
See `config.log' for more details
```

I couldn't tell if this was fatal or not so i just tried `./configure --disable-dependency-tracking` instead. That seemed to work, and I renamed the resulting `libhsl_ma86.so.0.0.0` to `libhsl.so` as per the instructions. And then got the error at the top when using it. Any ideas?

odow:
Closes #300 

cc @niclasmattsson was this all that was needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,8 +249,9 @@ _Tested on a clean install of Ubuntu 20.04._
 
 1. Install dependencies if necessary:
    ```
-   sudo apt-get install gfortran libblas-dev libmetis-dev
+   sudo apt install gfortran libblas-dev libmetis-dev
    ```
+   Note: on Windows Subsystem for Linux, you may also need `sudo apt install make`.
 2. Download the appropriate version of HSL.
    - MA27: [HSL for IPOPT from HSL](http://www.hsl.rl.ac.uk/ipopt/)
    - MA86: [HSL_MA86 from HSL](http://www.hsl.rl.ac.uk/download/HSL_MA86/1.6.0/)
--- a/README.md
+++ b/README.md
@@ -247,9 +247,9 @@ Currently untested. If you have instructions that work, please open an issue.
 
 _Tested on a clean install of Ubuntu 20.04._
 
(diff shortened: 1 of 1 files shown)


niclasmattsson:
Yes, assuming that `sudo apt-get install gfortran` does the same thing as `sudo apt install gfortran` (I did the latter). In my case I also needed to add the sudo to  `sudo make install`.

PR #313, opened 2022-04-07
DNMY: Exposed variable new_x in C_wrapper. Maintains backward comp.

sohailrreddy:
Updated C_wrapper to exposed the variable new_x from Ipopt routines to Julia functions. Maintains backward compatibility for cases where new_x is not used by routines.

### DIFF
--- a/src/C_wrapper.jl
+++ b/src/C_wrapper.jl
@@ -31,7 +31,7 @@ function _Eval_F_CB(
     if x_new == Cint(1)
         prob.x .= x
     end
-    new_x = convert(Bool, x_new)    
+    new_x = convert(Bool, x_new)
     new_obj = nothing
     if prob.expose_xnew
         new_obj = convert(Float64, prob.eval_f(x, new_x))::Float64
@@ -204,7 +204,7 @@ function CreateIpoptProblem(
     eval_grad_f,
     eval_jac_g,
     eval_h;
-    expose_xnew::Bool=false
+    expose_xnew::Bool = false,
 )
     @assert n == length(x_L) == length(x_U)
     @assert m == length(g_L) == length(g_U)
--- a/src/C_wrapper.jl
+++ b/src/C_wrapper.jl
@@ -30,7 +30,13 @@ function _Eval_F_CB(
     if x_new == Cint(1)
         prob.x .= x
     end
-    new_obj = convert(Float64, prob.eval_f(x))::Float64
+    new_x = convert(Bool, x_new)    
+    new_obj = nothing
+    try
+        new_obj = convert(Float64, prob.eval_f(x, new_x))::Float64
+    catch
+        new_obj = convert(Float64, prob.eval_f(x))::Float64
+    end
     unsafe_store!(obj_value, new_obj)
     return Cint(1)
 end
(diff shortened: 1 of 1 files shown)


odow (commented):
What's the purpose of this? 

We can't merge this as-is, because it's going to force all current models to throw an error on every function call, which will cause a massive performance problem.

Do you want to know if you need to recompute the function? You can do this in a closure.
```Julia
function memoize_f(f)
    last_x, last_fx = nothing, NaN
    return (x) -> begin
        if x == last_x
            return last_fx
        end
        last_x = x
        return f(x)
    end
end
```

sohailrreddy:
Understood, I updated it to remove the exceptions. The update shouldn't affect performance much and it should be faster to use Ipopt's xnew rather than recomputing the norm/checking equivalence outside every time eval_* are called.

odow:
What's the purpose of this? Do you have a benchmark where using a closure is a performance bottleneck?

src/C_wrapper.jl
```diff
@@ -30,23 +31,33 @@ function _Eval_F_CB(
     if x_new == Cint(1)
         prob.x .= x
     end
-    new_obj = convert(Float64, prob.eval_f(x))::Float64
+    new_x = convert(Bool, x_new)
```
odow, line 34: In the spirit of keeping closer to the C interface, why not just pass `x_new` in as the `Cint`?

src/C_wrapper.jl
```diff
@@ -30,23 +31,33 @@ function _Eval_F_CB(
     if x_new == Cint(1)
         prob.x .= x
     end
-    new_obj = convert(Float64, prob.eval_f(x))::Float64
+    new_x = convert(Bool, x_new)    
+    new_obj = nothing
+    if prob.expose_xnew
```
odow, line 36: You can also write
```Julia
new_obj = if prob.expose_xnew
    convert(Float64, ...)
else
    convert(Float64, ...)
end
```
That avoids the need for `new_obj = nothing` which creates a type instability.

src/C_wrapper.jl
```diff
@@ -169,7 +203,8 @@ function CreateIpoptProblem(
     eval_g,
     eval_grad_f,
     eval_jac_g,
-    eval_h,
+    eval_h;
+    expose_xnew::Bool=false
```
odow, line 207:
```suggestion
    expose_xnew::Bool = false,
```

src/C_wrapper.jl
```diff
@@ -83,23 +99,32 @@ function _Eval_Jac_G_CB(
...
 function _Eval_H_CB(
     n::Cint,
     x_ptr::Ptr{Float64},
-    ::Cint,
+    x_new::Cint,
     obj_factor::Float64,
     m::Cint,
     lambda_ptr::Ptr{Float64},
-    ::Cint,
+    new_lambda::Cint,
```
odow, line 127:
```suggestion
   ::Cint,
```
We don't name arguments if they aren't used. But if we're going th `x_new` route, should we not include `new_lambda` as well?

src/C_wrapper.jl
odow, line 34:
```suggestion
    new_x = convert(Bool, x_new)
```

sohailrreddy: On the Julia side, we might want to keep it closer to the Julia types?

src/C_wrapper.jl
sohailrreddy: Yes to be consistent, we should also expose new_lambda but I can't imagine many users using the exact hessian routines

src/C_wrapper.jl
odow: It's better to keep closer to the C types. Having differences between Julia and C is a pain (hence the reason you opened this PR).

odow:
I've updated this in https://github.com/jump-dev/Ipopt.jl/pull/322, and added some basic tests. I'd still like to see a benchmark where this makes a difference. You'd have to cache the old solutions, so why not just cache and check the `x` vector as well. It should be a pretty minor check in comparison to the full optimization, I can't see it being a bottleneck.

odow:
Closing in favor of #322. I'm still not convinced, but we can continue to discuss it in that PR. I'd like to see a practical application where this makes a difference.

PR #318, opened 2022-07-13
Fix support for MOI.Name

### LINKED ISSUE #317: add `MathOptInterface.Name()` attribute
When using Ipopt and ParametricOptInterface in direct mode this error happens: 

```Julia
julia> model = direct_model(POI.Optimizer(Ipopt.Optimizer()))
Error showing value of type Model:
ERROR: MathOptInterface.GetAttributeNotAllowed{MathOptInterface.Name}: Getting attribute MathOptInterface.Name() cannot be performed: Ipopt.Optimizer does not support getting the attribute MathOptInterface.Name(). You may want to use a `CachingOptimizer` in `AUTOMATIC` mode or you may need to call `reset_optimizer` before doing this operation if the `CachingOptimizer` is in `MANUAL` mode.
```

odow:
Closes #317

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -18,6 +18,7 @@ Create a new Ipopt optimizer.
 """
 mutable struct Optimizer <: MOI.AbstractOptimizer
     inner::Union{Nothing,IpoptProblem}
+    name::String
     invalid_model::Bool
     variables::MOI.Utilities.VariablesContainer{Float64}
     variable_primal_start::Vector{Union{Nothing,Float64}}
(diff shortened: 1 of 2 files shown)


blegat:
I don't think Ipopt should implement this

odow:
It's optional, but not unreasonable. We support other attributes that Ipopt itself doesn't (like `SolveTimeSec`).

blegat:
`SolveTimeSec` makes sense as we have no fallback for it and it's reasonable for users to want that.
However, I don't think we should complicate all solver wrappers with getter for the model if the solver does not support that.
We should make JuMP's printing use `try`-`catch` to be robust to that instead.

odow:
> We should make JuMP's printing use

Yeah, that was https://github.com/jump-dev/JuMP.jl/pull/3017. But I added it here in addition because it was trivial. Agree that not every solver has to implement.

PR #321, opened 2022-07-17
Rebuild IpoptProblem only if modified

odow:
Part of https://github.com/jump-dev/JuMP.jl/issues/1185
x-ref https://github.com/jump-dev/JuMP.jl/pull/3018

For @ccoffrin's PF example: https://github.com/lanl-ansi/nlp-jump-examples/pull/9 (private link at present) 

Without this PR
```
julia> @time bench_pf_nlparameter(case="pglib_opf_case2000_goc.m", n = 100)
 83.591282 seconds (225.00 M allocations: 34.441 GiB, 14.97% gc time, 1.07% compilation time)
```

With this PR
```
julia> @time bench_pf_nlparameter(case="pglib_opf_case2000_goc.m", n = 100)
 21.147063 seconds (27.64 M allocations: 1.517 GiB, 4.47% gc time)
```

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -25,6 +25,7 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     variable_lower_start::Vector{Union{Nothing,Float64}}
     variable_upper_start::Vector{Union{Nothing,Float64}}
     nlp_data::MOI.NLPBlockData
+    nlp_data_needs_initialize::Bool
     sense::MOI.OptimizationSense
     objective::Union{
         Nothing,
@@ -81,6 +82,7 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
             Union{Nothing,Float64}[],
             Union{Nothing,Float64}[],
             MOI.NLPBlockData([], _EmptyNLPEvaluator(), false),
+            true,
             MOI.FEASIBILITY_SENSE,
             nothing,
             _ConstraintInfo{
@@ -693,6 +695,7 @@ MOI.supports(::Optimizer, ::MOI.NLPBlock) = true
 
 function MOI.set(model::Optimizer, ::MOI.NLPBlock, nlp_data::MOI.NLPBlockData)
     model.nlp_data = nlp_data
+    model.nlp_data_needs_initialize = true
     return
 end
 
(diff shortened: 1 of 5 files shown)


ccoffrin:
A nice improvement for such minor updates.

odow:
It's actually not so straightforward, because it relies on https://github.com/jump-dev/JuMP.jl/pull/3018#issuecomment-1186677741. And it isn't obvious whether we need to call `MOI.initialize` if a parameter changes.

odow:
Regardless of the JuMP change, this is still a win if you want to repeatedly solve the same problem with different primal/dual starting points.

src/MOI_wrapper.jl
```diff
@@ -231,7 +233,9 @@ function MOI.get(
 end
 
 function MOI.add_constraint(model::Optimizer, x::MOI.VariableIndex, set::_SETS)
-    return MOI.add_constraint(model.variables, x, set)
+    index = MOI.add_constraint(model.variables, x, set)
+    model.inner = nothing
```
frapac, line 237: Just a quick thought: suppose we have solved a problem before, and `model.inner != nothing`. If we add a variable or a constraint we delete the model previously built by setting `model.inner = nothing` here. 

Rather than doing this implicitly, shouldn't we issue a warning to the user (explicit being better than implicit)? Or do we consider that deleting the model is a minor operation not worth reporting?

odow: The `.inner` is an implementation detail. I think we're fine to clear it internally.

src/MOI_wrapper.jl
```diff
@@ -231,7 +233,8 @@ function MOI.get(
...
```
odow, line 237:
```suggestion
    model.inner = nothing
    return index
```

odow:
This does't solve the JuMP issue, because that needs changes in MOI/JuMP, but this is an uncontroversial change in Ipopt that still improves things somewhat (we don't rebuild if you want to change the starting point and re-optimize).

ccoffrin:
Still there are some good uses cases of this.  Like multi-start inside of MINLP algorithms.

odow:
> Like multi-start inside of MINLP algorithms.

Yes, this is exactly what I had in mind.

PR #322, opened 2022-07-18
RFC: Expose xnew in C_wrapper, maintaining compat

odow:
Replacement for #313, just so I can try it out. I don't know if we should merge this. The alternative is for the user to write a closure, which isn't that much work.

cc @sohailrreddy

### DIFF
--- a/src/C_wrapper.jl
+++ b/src/C_wrapper.jl
@@ -21,6 +21,7 @@ mutable struct IpoptProblem
     eval_jac_g::Function
     eval_h::Union{Function,Nothing}
     intermediate::Union{Function,Nothing}
+    expose_xnew::Bool
 end
 
 function _Eval_F_CB(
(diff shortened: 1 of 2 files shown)


odow:
@sohailrreddy are you still interested in this? If not, I will close.

From https://github.com/jump-dev/Ipopt.jl/pull/313#issuecomment-1217459540:

> I'd like to see a practical application where this makes a difference.

odow:
Closing as stale. I understand the reasoning for this, so if anyone has a benchmark where this makes a difference in practice, please post a comment and we can re-open the PR.

DaniFojo:
Hello. Exposing xnew could really be useful for us. 

The reason is that our methods to compute objective, grad and constraint values all need to compute some shared data first, and then use that shared data. If `xnew` is exposed we can know if we need to recompute that shared data. The memoization closure won't work for us.

For example, if `eval_f` was just called with a given x, if `eval_grad_f` is called with the same x we can skip this computation.

Right now we are using our own fork of the package to do this, but I think it could be update this PR and merge it.

PR #327, opened 2022-08-24
Ipopt_jll with LBT

### LINKED ISSUE #6: distribute with MKL?
@ViralBShah mentioned in JuliaLang/julia#4272 that julia has a license to redistribute MKL. Could we use this for the Ipopt binaries? This should give much better default performance than with MUMPS.

@tkelman

amontoison:
close #6 

I updated and recompiled `MUMPS_seq_jll` / `Ipopt_jll` with [libblastrampoline](https://github.com/JuliaLinearAlgebra/libblastrampoline).
We can easily switch the BLAS backend now with the version `300.1400.401` of `Ipopt_jll`.
```
import LinearAlgebra, OpenBLAS32_jll
LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path)
```
or
```
import LinearAlgebra, MKL_jll
LinearAlgebra.BLAS.lbt_forward(MKL_jll.libmkl_rt_path)
```
Because `Ipopt_jll` and `MUMPS_seq_jll` are compiled with `LBT_jll` now, we must add OpenBLAS32_jll as dependency here to have a least one 32 bit BLAS/LAPACK backend.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 [compat]
 Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.401"
 MathOptInterface = "1"
-OpenBLAS32_jll = "0.3.10"
+OpenBLAS32_jll = "=0.3.20"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -12,6 +12,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 [compat]
 Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.401"
 MathOptInterface = "1"
+OpenBLAS32_jll = "0.3.10"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 [compat]
 Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.401"
 MathOptInterface = "1"
-OpenBLAS32_jll = "=0.3.20"
+OpenBLAS32_jll = "0.3.10"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 [compat]
 Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.401"
 MathOptInterface = "1"
-OpenBLAS32_jll = "0.3.10"
+OpenBLAS32_jll = "=0.3.10"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 [compat]
 Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.401"
 MathOptInterface = "1"
-OpenBLAS32_jll = "=0.3.10"
+OpenBLAS32_jll = "0.3.10"
 julia = "1.6"
 
 [extras]
--- a/Project.toml
+++ b/Project.toml
@@ -5,10 +5,12 @@ version = "1.0.4"
 
 [deps]
 Ipopt_jll = "9cc047cb-c261-5740-88fc-0cf96f7bdcc7"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 
 [compat]
-Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400"
+Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.401"
 MathOptInterface = "1"
 julia = "1.6"
 
--- a/Project.toml
+++ b/Project.toml
@@ -10,7 +10,7 @@ MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 
 [compat]
-Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.401"
+Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.402"
 MathOptInterface = "1"
 OpenBLAS32_jll = "0.3.10"
 julia = "1.6"
--- a/Project.toml
+++ b/Project.toml
@@ -10,7 +10,7 @@ MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 
 [compat]
-Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.402"
+Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.403"
 MathOptInterface = "1"
 OpenBLAS32_jll = "0.3.10"
 julia = "1.6"
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -14,7 +14,10 @@ const MOI = MathOptInterface
 
 function __init__()
     if VERSION >= v"1.8"
-        LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path; verbose = true)
+        LinearAlgebra.BLAS.lbt_forward(
+            OpenBLAS32_jll.libopenblas_path;
+            verbose = true,
+        )
     end
     global libipopt = Ipopt_jll.libipopt
     return
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -6,11 +6,14 @@
 module Ipopt
 
 import Ipopt_jll
+import OpenBLAS32_jll
 import MathOptInterface
+import LinearAlgebra
 
 const MOI = MathOptInterface
 
 function __init__()
+    (VERSION ≥ v"1.8") && LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path)
     global libipopt = Ipopt_jll.libipopt
     return
 end
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -13,7 +13,9 @@ import LinearAlgebra
 const MOI = MathOptInterface
 
 function __init__()
-    (VERSION ≥ v"1.8") && LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path)
+    if VERSION >= v"1.8"
+        LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path)
+    end
     global libipopt = Ipopt_jll.libipopt
     return
 end
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -14,7 +14,7 @@ const MOI = MathOptInterface
 
 function __init__()
     if VERSION >= v"1.8"
-        LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path)
+        LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path; verbose = true)
     end
     global libipopt = Ipopt_jll.libipopt
     return
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -13,7 +13,7 @@ import OpenBLAS32_jll
 const MOI = MathOptInterface
 
 function __init__()
-    if VERSION >= v"1.8"
+    if false # VERSION >= v"1.8"
         config = LinearAlgebra.BLAS.lbt_get_config()
         if !any(lib -> lib.interface == :lp64, config.loaded_libs)
             LinearAlgebra.BLAS.lbt_forward(
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -13,12 +13,13 @@ import OpenBLAS32_jll
 const MOI = MathOptInterface
 
 function __init__()
-    if false # VERSION >= v"1.8"
+    if VERSION >= v"1.8"
         config = LinearAlgebra.BLAS.lbt_get_config()
         if !any(lib -> lib.interface == :lp64, config.loaded_libs)
             LinearAlgebra.BLAS.lbt_forward(
                 OpenBLAS32_jll.libopenblas_path;
                 verbose = true,
+                clear = false,
             )
         end
     end
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -6,18 +6,21 @@
 module Ipopt
 
 import Ipopt_jll
-import OpenBLAS32_jll
-import MathOptInterface
 import LinearAlgebra
+import MathOptInterface
+import OpenBLAS32_jll
 
 const MOI = MathOptInterface
 
 function __init__()
     if VERSION >= v"1.8"
-        LinearAlgebra.BLAS.lbt_forward(
-            OpenBLAS32_jll.libopenblas_path;
-            verbose = true,
-        )
+        config = LinearAlgebra.BLAS.lbt_get_config()
+        if !any(lib -> lib.interface == :lp64, config.loaded_libs)
+            LinearAlgebra.BLAS.lbt_forward(
+                OpenBLAS32_jll.libopenblas_path;
+                verbose = true,
+            )
+        end
     end
     global libipopt = Ipopt_jll.libipopt
     return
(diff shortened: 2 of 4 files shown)


Project.toml
```diff
@@ -5,10 +5,12 @@ version = "1.0.4"
 
 [deps]
 Ipopt_jll = "9cc047cb-c261-5740-88fc-0cf96f7bdcc7"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
```
odow, line 10: This needs a compat bound

src/Ipopt.jl
```diff
@@ -6,11 +6,14 @@
...
 
 import Ipopt_jll
+import OpenBLAS32_jll
 import MathOptInterface
+import LinearAlgebra
 
 const MOI = MathOptInterface
 
 function __init__()
+    (VERSION ≥ v"1.8") && LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path)
```
odow, line 16:
```suggestion
    if VERSION >= v"1.8"
        LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path)
    end
```

odow:
Can we test with MKL on CI machines?

amontoison:
> Can we test with MKL on CI machines?

Yes, you just need to add `MKL_jll` as a test dependency.
This simple modification should work.
```
@testset "C" begin
    include("C_wrapper.jl")
end

@testset "MathOptInterface" begin
    include("MOI_wrapper.jl")
end

import LinearAlgebra, MKL_jll
LinearAlgebra.BLAS.lbt_forward(MKL_jll.libmkl_rt_path)

@testset "C" begin
    include("C_wrapper.jl")
end

@testset "MathOptInterface" begin
    include("MOI_wrapper.jl")
end
```

Project.toml
amontoison: Isn't already done during the compilation in Yggdrasil?
Each `OpenBLAS32_jll` has a julia compat.

odow: All dependencies (excluding stdlibs) need a compat bound in the package that declares them as a dependency.

odow:
Windows segfault seems related: https://github.com/jump-dev/Ipopt.jl/runs/7986459839?check_suite_focus=true

Does LBT not work on Windows?

amontoison:
> Windows segfault seems related: https://github.com/jump-dev/Ipopt.jl/runs/7986459839?check_suite_focus=true
> 
> Does LBT not work on Windows?

I never tested LBT on Windows. I don't know if it's related to the new version of MUMPS (5.5.1) or LBT.
I also use LBT for HSL.jl but Windows was not tested.

amontoison:
> Windows segfault seems related: https://github.com/jump-dev/Ipopt.jl/runs/7986459839?check_suite_focus=true
> 
> Does LBT not work on Windows?

It seems to be related to OpenBLAS32.
We should test with `v0.3.10` and `v0.3.12` to see if we also have this error.

odow:
The null pointer suggests that something isn't loaded properly. 

@ViralBShah is there anything we need to do for LBT on Windows?

ViralBShah:
Not to the best of my knowledge.

ViralBShah:
I suggest passing all the arguments to lbt_forward to make sure we are not clearing the ilp64 symbols.

ViralBShah:
Also why such an old OpenBLAS32?

ViralBShah:
Also, perhaps we want to only forward if the user has not already set up a forward. Otherwise you have to load packages in the right order. Eventually we need a better Julia wide mechanism to use Preferences to pick BLAS etc.

Project.toml
```diff
@@ -5,11 +5,14 @@ version = "1.0.4"
...
 
 [compat]
-Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400"
+Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.401"
 MathOptInterface = "1"
+OpenBLAS32_jll = "=0.3.10"
```
odow, line 15:
```suggestion
OpenBLAS32_jll = "0.3.10"
```

src/Ipopt.jl
```diff
@@ -6,11 +6,16 @@
...
 import Ipopt_jll
+import OpenBLAS32_jll
 import MathOptInterface
+import LinearAlgebra
 
 const MOI = MathOptInterface
 
 function __init__()
+    if VERSION >= v"1.8"
+        LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path)
```
odow, line 17:
```suggestion
        LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path; verbose = true)
```

amontoison:
> Also, perhaps we want to only forward if the user has not already set up a forward. Otherwise you have to load packages in the right order so that MKL.jl forwards are not overwritten by this package for example. Eventually we need a better Julia wide mechanism to use Preferences to pick BLAS etc.

Is it possible to detect if the user has already set up a 32 bits BLAS/LAPACK backend ?

ViralBShah:
> Is it possible to detect if the user has already set up a 32 bits BLAS/LAPACK backend ?

Yes, LBT allows you to query with `BLAS.lbt_get_config()`

ViralBShah:
> Does LBT not work on Windows?

Julia base uses LBT by default, and so this is checked in Julia base CI etc.

giordano:
> is there anything we need to do for LBT on Windows?

I'm not aware of anything special being needed for Windows

ViralBShah:
Can someone fix the CI and re-trigger?

odow:
I added a check for LP64 BLAS, but the problem is still there.

odow:
Does it matter that OpenBLAS32_jll and OpenBLAS_jll are different versions?

ViralBShah:
It shouldn't really matter, but one can never be 100% sure.

ViralBShah:
Let's test with 0.3.20 and see if that helps?

odow:
Ah. 3.20 isn't supported on Julia v1.8

odow:
So even when we don't forward, Windows still hits the same issue. Perhaps this is unrelated to LBT?

ViralBShah:
The messages in the log files suggest that the BLAS/LAPACK are not being loaded on 1.8. I'm taking a look.

ViralBShah:
We have the same version of LBT in 1.8 and nightly. So I wonder why 1.8 is failing and nighty is passing. @giordano Any ideas?

ViralBShah:
Ah I didn't realize that the `lbt_forward` was currently disabled with the `if false` flag. Turning it on, the tests go through fine, although I have yet to test on Windows. It is good to set `clear = false` for `lbt_forward` as well - I believe it is the default, but best to be deliberate about it.

odow:
> Ah I didn't realize that the lbt_forward was currently disabled with the if false flag

I disabled it to check if we get the same fault without forwarding. We do, so I wonder if the problem lies elsewhere?

I've re-enabled with `clear = false`

ViralBShah:
Yes, I see what you were doing. It suggests the problem is perhaps elsewhere with Ipopt. Do we know if this version of Ipopt works on windows? Perhaps we should try with an earlier version?

I'll try to see if I can reproduce on a Windows machine.

ViralBShah:
It says it crashes in line 219 in C_wrapper tests, which is just the function call to the tests. Can we run the tests temporarily in a script or something - with the hope that we get a more direct failing test case?

odow:
CI was passing a few days ago on Julia v1.7 with these versions:
image

I just kicked off #328 to see if it's a Julia v1.8 thing.

I'll modify the test script for this PR

amontoison:
> Yes, I see what you were doing. It suggests the problem is perhaps elsewhere with Ipopt. Do we know if this version of Ipopt works on windows? Perhaps we should try with an earlier version?
> 
> I'll try to see if I can reproduce on a Windows machine. Something is clearly giving it a null pointer.

For Julia 1.6 or 1.7, we have Ipopt 3.14.4 + MUMPS_seq 5.4.1 + OpenBLAS32.
For Julia 1.8, we have Ipopt 3.14.4 + MUMPS_seq 5.5.1 + LBT.

If the issue comes from MUMPS_seq, we never tested the version 5.5.1 on Windows.

odow:
That's weird. The issue is in `CreateIpoptProblem`. 

> For Julia 1.6 or 1.7, we have Ipopt 3.14.4 + MUMPS_seq 5.4.1 + OpenBLAS32.

Also work on Julia v1.8 with those versions.

ViralBShah:
Well it just means that the corruption is happening somewhere earlier or in there. Since we are seeing a null pointer, I am guessing something bad is happening earlier. Worth doing a quick audit of the ccall signatures and ideally, once the msan infrastructure in julia is ready, we can build Ipopt and dependencies with msan.

As @amontoison points out, the issue could also just be in any of the dependencies like MUMPS or LBT+BLAS.

odow:
The good news is that nothing happens earlier. This is the first call to Ipopt. I also checked the ccalls. They look correct. The interface has been pretty well tested, and we didn't change the version of Ipopt. The only change was MUMPS and LBT.

ViralBShah:
I suppose the next thing is to try out some other less complex library with LBT and OpenBLAS32, and test MUMPS_seq_jll too.

amontoison:
With @dpo we updated `MUMPS.jl`, it should not too difficult to test `MUMPS_seq_jll` instead of `MUMPS_jll`.

ViralBShah:
Since the first test is failing, it suggests that something is broken with the build. I suppose one intermediate step is to bump the dependencies without using LBT first, to make sure those work. If so, we know LBT is the culprit, and we need to track it down.

amontoison:
The issue is no related to `Ipopt_jll` / `Ipopt.jl`.
It is LBT or `MUMPS_jll`.
https://github.com/JuliaSmoothOptimizers/MUMPS.jl/runs/8027204285?check_suite_focus=true#step:6:211

dpo:
MUMPS_seq_jll?

ViralBShah:
What's the easiest way to test the new MUMPS_seq_jll? `Pkg.test("MUMPS.jl")` will try to do the whole MPI testing, right?

amontoison:
@ViralBShah You just need a branch to use the branch `mumps_seq`.
I replaced `MUMPS_jll` by `MUMPS_seq_jll`.
```julia
julia> ]
pkg> add MUMPS#mumps_seq
pkg> test MUMPS
```

odow:
I just had a chat to @amontoison and our plan is:

 * Recompile `MUMPS_seq_jll` v5.4.1 as v500.400.100 with LBT
 * Recompile Ipopt to use MUMPS_seq_jll@500.400.100

Because we know that Ipopt works with MUMPS v5.4.1, this should let us know whether the problem is LBT or MUMPS. Upgrading from 5.4.1 to 5.5.1 and from OpenBLAS to LBT at the same time confused things.

ViralBShah:
In the next few days, I'll also actually find a windows machine to test things.

amontoison:
I recompiled `MUMPS_seq`: https://github.com/JuliaPackaging/Yggdrasil/pull/5393

ViralBShah:
With MUMPS#mumps_seq, when I try to `dlopen` the `libcmumps.dll` file, I get `The specified module could not be found`.

giordano:
Use `Library("C:\\Users\\viral\\.julia\\artifacts\\029cea9e5bf8cca60ec7023a09ff1c9a45091f9a\\bin\\libcmumps.dll")` from https://github.com/giordano/DependencyWalker.jl in the same session where loading the library fails

ViralBShah:
Seems like an issue with LBT. Perhaps not built correctly? https://github.com/JuliaPackaging/Yggdrasil/pull/5309

```
julia> Library("C:\\Users\\viral\\.julia\\artifacts\\029cea9e5bf8cca60ec7023a09ff1c9a45091f9a\\bin\\libcmumps.dll")
◼ C:\Users\viral\.julia\artifacts\029cea9e5bf8cca60ec7023a09ff1c9a45091f9a\bin\libcmumps.dll
  ✗ libmumps_common.dll (NOT FOUND)
  ✗ libmpiseq.dll (NOT FOUND)
  ✗ libmetis.dll (NOT FOUND)
  ◼ C:\WINDOWS\System32\msvcrt.dll
    ◼ C:\WINDOWS\System32\KERNELBASE.dll
  ✗ libblastrampoline-5-0-2.dll (NOT FOUND)
  ◼ C:\Users\viral\.julia\juliaup\julia-1.8.0+0.x64\bin\libgfortran-5.dll
    ◼ C:\Users\viral\.julia\juliaup\julia-1.8.0+0.x64\bin\libwinpthread-1.dll
      ◼ C:\WINDOWS\System32\msvcrt.dll
        ◼ C:\WINDOWS\System32\KERNELBASE.dll
      ◼ C:\WINDOWS\System32\KERNEL32.DLL
        ◼ C:\WINDOWS\System32\KERNELBASE.dll
```

giordano:
What version of Julia are you using?  I'm very worried by that `5-0-2` in the name of the library, why there is the full version number?  That looks a terrible idea, the kind which prevents ABI-compatibility across different versions.

ViralBShah:
Julia 1.8.0.

giordano:
Where should `libcmumps.dll` come from?  You have to run `Library` _after_ you have attempted to load all the dependencies, otherwise Julia can't possibly find them.

giordano:
There are some issues with LBT on Windows: https://github.com/JuliaLinearAlgebra/libblastrampoline/issues/87

amontoison:
I confirm that we have also the null pointer with MUMPS 5.4.1 + LBT.
https://github.com/JuliaSmoothOptimizers/MUMPS.jl/runs/8124551353?check_suite_focus=true#step:6:208

ViralBShah:
Can we also try 32-bit Windows here?

odow:
> Can we also try 32-bit Windows here?

Added. Running now: https://github.com/jump-dev/Ipopt.jl/runs/8148650553?check_suite_focus=true

odow:
Same null pointer error on 32-bit

amontoison:
@odow, I rebuild `MUMPS_seq` and `Ipopt`. I hope that it works now :crossed_fingers:

test/C_wrapper.jl
```diff
@@ -3,12 +3,12 @@
 # Use of this source code is governed by an MIT-style license that can be found
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
-module TestCWrapper
+# module TestCWrapper
```
odow, line 6: Looks like you included these changes by mistake?

test/C_wrapper.jl
```diff
@@ -211,7 +211,7 @@ function test_hs071()
     finalize(prob) # Needed before the `rm` on Windows.
     # unlink the output file
     rm("blah.txt")
-    return
+   return
```
odow, line 214:
```suggestion
    return
```

odow:
@amontoison happy? So we've just disabled LBT on Windows for now?

amontoison:
> @amontoison happy? So we've just disabled LBT on Windows for now?

Yes, it's finally working!
We will not use LBT on Windows until Julia v"1.9". Like OpenBLAS the version of LBT is fixed for each Julia version.
With Julia v"1.9", the issues with Windows should be fixed and we will be able to recompile again MUMPS and Ipopt (if we want to support the same feature on Windows).

@ViralBShah Is it possible to do a new release of `MKL.jl` such that `using MKL` automatically load a LP64 BLAS?
We already have this feature but we must use the branch `master` for now.

For Apple platforms, we could use directly Apple Accelerate what do you think?
We could have a significant speed-up with the new M1 architecture.
It's a LP64, which is exactly what we need for MUMPS / Ipopt.  
```julia
if VERSION >= v"1.8"
    config = LinearAlgebra.BLAS.lbt_get_config()
    if !any(lib -> lib.interface == :lp64, config.loaded_libs)
        if !Sys.isapple()
            LinearAlgebra.BLAS.lbt_forward(OpenBLAS32_jll.libopenblas_path, verbose = true, clear = false)
        else
            blas = "/System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/libBLAS.dylib"
            lapack = "/System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/libLAPACK.dylib"
            LinearAlgebra.BLAS.lbt_forward(blas, verbose = true, clear = false)
            LinearAlgebra.BLAS.lbt_forward(lapack, verbose = true, clear = false)
        end
    end
end
```

ViralBShah:
I have been holding off on a new release because IIRC, the latest master does not pass on one of the windows. 

I will trigger another build today. On Mac it does make sense to link to MKL.

amontoison:
MKL is not efficient on Mac because it's only working with one thread:
https://github.com/JuliaLinearAlgebra/MKL.jl/issues/111

ViralBShah:
Oops - sorry I was typing on my phone in a hurry. I meant to say that on mac, it does make sense to link to Accelerate.

src/Ipopt.jl
```diff
@@ -6,11 +6,23 @@
...
 
 import Ipopt_jll
+import LinearAlgebra
 import MathOptInterface
+import OpenBLAS32_jll
 
 const MOI = MathOptInterface
 
 function __init__()
+    if VERSION >= v"1.8"
```
odow, line 16: Do we need to check if we are using `Ipopt_jll@300.1400.403`? What are the implications of running this block if we are using a non-LBT version of Ipopt_jll?

ViralBShah:
I've tagged MKL 0.6

ViralBShah:
If 32-bit BLAS is being linked against through LBT, then on Apple, we should also be able to set up forwarding to Accelerate and see if performance is better.

src/Ipopt.jl
amontoison: It will load a 32 bits BLAS / LAPACK but it's not harmful.

amontoison:
Thanks for the new release of MKL @ViralBShah. For Apple Accelerate, we did some tests with @dpo for HSL and the performances seem similar.
https://github.com/JuliaSmoothOptimizers/HSL.jl/pull/124#issuecomment-1272589104
 We decided to use OpenBLAS32 by default and add documentation to explain how to load MKL or Apple Accelerate.

odow:
@amontoison are you happy for this to be merged?

amontoison:
Yes, It's a nice feature for Ipopt.jl.

Issue #329, opened 2022-08-26
Allow user to specify the Ipopt binary


etatara:
I'm updating a project from Julia 1.6 and Ipopt.jl 0.9.1 to Julia 1.8 and Ipopt.jl 1.0.4 and am wondering if it is still possible to specify a custom Ipopt binary as before using JULIA_IPOPT_LIBRARY_PATH?  I discovered that using the Ipopt binary (3.14.4) that's installed automatically with Ipopt.jl 1.0.4 works fine with the HSL linear solvers, but it runs significantly slower and with more iterations than the custom built Ipopt library (3.14.5) I made with HSL.  I can confirm this by changing the symlink in the ipopt artifact to point to my custom built library.

Considering that platform-dependent performance may be an important issue with Ipopt, it would be great to be able to easily specify the path to a custom binary.

odow:
See https://jump.dev/JuMP.jl/stable/developers/custom_solver_binaries/

etatara:
> See https://jump.dev/JuMP.jl/stable/developers/custom_solver_binaries/

That is really an elegant solution!  Thanks very much.  I am continuously impressed by the maturation of the jump-dev codebase.  I would humbly suggest adding this short bit of critical information to the Ipopt.jl readme.

PR #330, opened 2022-08-28
[README] add link to custom solver binaries

odow:
cc @etatara 

x-ref #329


blegat (approved):

blegat:
We could add that to other solvers as well

Issue #339, opened 2022-10-25
Problem with HSL solvers since updating to MacOS 13.0


this-josh:
I installed the HSL solvers while on MacOS 12.x and they worked fine, now on MacOS 13.0 I'm finding the following:
1. Launching Julia from the HSL library path (/Users/jak121/CoinOR/HSL/lib) gives me access to the HSL linear solvers
2. Using `set_optimizer_attribute(model, "hsllib", "/Users/jak121/CoinOR/HSL/lib/libhsl.dylib")` works
3. Running`export DL_LOAD_PATH=/Users/jak121/CoinOR/HSL/lib` and moving to a separate dir gives
```shell
Exception of type: DYNAMIC_LIBRARY_FAILURE in file "Common/IpLibraryLoader.cpp" at line 67:
 Exception message: dlopen(libhsl.dylib, 0x0002): tried: 'libhsl.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibhsl.dylib' (no such file), '/Users/jak121/.julia/artifacts/5f87ee5a0999062b66c1db490761d65343e0613f/lib/./libhsl.dylib' (no such file), '/Users/jak121/.julia/juliaup/julia-1.8.2+0.aarch64/lib/libhsl.dylib' (no such file), '/usr/lib/libhsl.dylib' (no such file, not in dyld cache), 'libhsl.dylib' (no such file)

EXIT: Library loading failure.
```
which suggests to me that `DL_LOAD_PATH` isn't the correct environment variable to set as it doesn't even seem to be trying this path. I have also tried setting `LD_LIBRARY_PATH` and this does not work. 

After step 3 if I run Julia I get
```Julia
julia> using Libdl

julia> dlopen("sys")
Ptr{Nothing} @0x00000000a1e14350

julia> dlopen("libhsl")
ERROR: could not load library "libhsl"
dlopen(libhsl.dylib, 0x0001): symbol not found in flat namespace '_metis_nodend_'
Stacktrace:
 [1] dlopen(s::String, flags::UInt32; throw_error::Bool)
   @ Base.Libc.Libdl ./libdl.jl:117
 [2] dlopen (repeats 2 times)
   @ ./libdl.jl:116 [inlined]
 [3] top-level scope
   @ REPL[3]:1

julia> using JuMP,Ipopt

julia> begin
           m = Model(optimizer_with_attributes(Ipopt.Optimizer, "linear_solver"=>"ma27"))
[...]

PR #340, opened 2022-10-27
Updated HSL linear solvers

this-josh:
- Added some points on WSL
- Updated MacOS instructions closing #339

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ _Tested on a clean install of Ubuntu 20.04._
 
 #### Mac
 
-_Tested on a MacBook Pro, 10.15.7._
+_Tested on a MacBook Pro, 10.15.7, 12.6, 13.0_
 
 1. Download the appropriate version of HSL.
    - MA27: [HSL for IPOPT from HSL](http://www.hsl.rl.ac.uk/ipopt/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -300,6 +300,7 @@ _Tested on a MacBook Pro, 10.15.7._
    - For `ma86`, the file is `/full/path/somewhere/lib/libhsl_ma86.dylib`
 4. Place the `libhsl.dylib` library somewhere on your load path.
    - Alternatively, start Julia with `export DL_LOAD_PATH=/full/path/somewhere/lib; julia`
+     - If this doesn’t work try `set_optimizer_attribute(model, "hsllib","full/path/somewhere/lib/libhsl.dylib")`
```
odow, line 303: Is it simpler if we replace steps 3, 4, and 5 with something like this?
````
 3. Set the `hsllib` and `linear_solver` options to the full path to the library
    and the name of the solver, respectively. For example:
    ```julia
    using JuMP, Ipopt
    set_optimizer_attribute(
        model,
        "hsllib", 
        "/full/path/to/somewhere/lib/libhsl_ma86.dylib",
    )
    set_optimizer_attribute(model, "linear_solver", "ma86")
    # or
    set_optimizer_attribute(
        model,
        "hsllib", 
        "/full/path/to/somewhere/lib/libcoinhsl.dylib",
    )
    set_optimizer_attribute(model, "linear_solver", "ma27")
    ```
````

There's also `pardisolib` https://coin-or.github.io/Ipopt/OPTIONS.html#OPT_pardisolib for the pardiso section

this-josh: Perhaps but I'm not sure whether my `DL_LOAD_PATH` issues are anomalous or the norm, so I didn't want to remove the previous content. For me `libhsl.dylib` is able to do both `ma27` and `ma86` so there's no need to differentiate. Unfortunately, I don't have access to `pardiso` so I'm unable to verify any solutions.

odow:
Thanks!

PR #343, opened 2022-11-22
Change cpu_time to wall_time

LukasBarner:
While setting up the fork for updating the docs on linear solvers, I remembered that this was a thing that sometimes bugged me a lot: 
When using different linear solvers, cpu time can massively deviate from wall time. For example, when using pardiso on a cluster computer, I have had instances that exited with a cpu time of 10000s after less than 400s wall time. 
The behavior varies widely between linear solvers, which makes things somewhat inconsistent. 
If this does not violate JuMP/MOI conventions (think the MOI docs are not really explicit on this), it might be more user-friendly to go to wall time for `MOI.TimeLimitSec`? 
Users who need cpu time could still manually set the option (as is currently possible for wall time).  
In case this is a stupid idea for another reason, please just delete :D

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,7 +10,7 @@ MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 
 [compat]
-Ipopt_jll = "=3.13.2, =3.13.4, =300.1400.400, =300.1400.403"
+Ipopt_jll = "=300.1400.400, =300.1400.403"
 MathOptInterface = "1"
 OpenBLAS32_jll = "0.3.10"
 julia = "1.6"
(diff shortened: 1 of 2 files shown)


odow:
The main reason was that this was introduced in v3.14.0 of Ipopt.

We'd have to also drop support for 3.13 here (which is okay, I guess, the time limit problem is annoying):
https://github.com/jump-dev/Ipopt.jl/blob/292640b936e76bcacddbe7841c99526054f9d40e/Project.toml#L13

mlubin:
Agreed wall time is a better (less surprising) limit than cpu time.

blegat:
> cpu time of 10000s after less than 400s wall time.

Isn't the wall time supposed to be larger than the cpu time ? Or is that because it adds up the cpu time of different processes ?

LukasBarner:
> Isn't the wall time supposed to be larger than the cpu time ? Or is that because it adds up the cpu time of different processes ? 

Yes, I think for parallel linear solvers time is added up. Due to their different design, that is also why they produce different wall time results for the same cpu time limit.

blegat:
In that case, the wall time is better indeed

blegat (approved):

Issue #342, opened 2022-11-22
Linking full hsl library to Ipopt.jl on ubuntu


bennerh:
Hi everyone,

I tried to add the full hsl library to Ipopt on a clean ubuntu following the ubuntu specific instructions on from [here](https://github.com/jump-dev/Ipopt.jl#custom-installation). However, running a basic optimization problem

```
model = Model(Ipopt.Optimizer)
set_optimizer_attribute(model, "linear_solver", "ma27")
@variable(model, 0 <= x <= 2)
@variable(model, 0 <= y <= 30)
@objective(model, Max, 5x + 3 * y)
@constraint(model, con, 1x + 5y <= 3)
optimize!(model)
```

 Leads to the following error message

```
Exception of type: DYNAMIC_LIBRARY_FAILURE in file "Common/IpLibraryLoader.cpp" at line 67:
 Exception message: <path/to/libhsl.so>: undefined symbol: dtrmm_
```

Other linear solver attributes as "ma57" lead to the same message.

I tried installation on both Ubuntu 18.04 and 20.04, using Ipopt version 1.1.0 and JuMP version 1.4.0

Are there any ideas of what I might have forgotten? Thanks a lot in advance!

LukasBarner:
You could try the following: 
```julia
using Libdl
Libdl.dlopen("ENTER_PATH_TO_liblapack.so.3", RTLD_GLOBAL)
```
prior to your code snippet and do the same for following errors (i.e. if symbols related to other libraries are undefined). Not sure if that's the best/recommended way, but I think it might do the trick.

odow:
Compiling the HSL libraries and getting everything linked can be finicky. If you have improvements to the instructions, please open a PR.

LukasBarner:
I will make a PR, but let's see first if this works out :D

amontoison:
@bennerh 
I had the exact same problem this afternoon when I tried to use the `configure` script of `coinhsl`.
I just released a new version of HSL.jl (0.3.5) to compile it automatically.

LukasBarner:
Do any of you see problems with this approach?
If not, I could make a PR to include something like this in the instructions.

LukasBarner:
At least on an hpc cluster, I encountered the same problem despite metis and lapack being on the load path.
just running:
```julia
dlopen("ENTER_PATH_TO_liblapack.so.3", RTLD_GLOBAL)
```
did the trick for me. 
[...]

Issue #341, opened 2022-11-21
double free or corruption (out) error


LukasBarner:
Hi everyone, 
I am solving a series of MPECs, where a relaxation on complementary slackness conditions is tightened every iteration. The model code works as expected most of the time, however i do sometimes get strange memory related errors with Ipopt and MA97. Likely, they are not related to the julia interface, but I think it might be best to start here and make my way downstream. I have attached a log of the last unsuccessful iteration below. 
The error is likely related to some numerically odd constellation, as the 20 successful runs before that did use the same code and only had different values. 
Do you think there is a better way to dig into this than just by creating a more detailed log?
If not, I will try to come up with a more detailed report, but this might take some time as the model needs to run all previous iterations again...

```
This is Ipopt version 3.14.4, running with linear solver ma97.


Detected 9770 linearly dependent equality constraints; taking those out.

Number of nonzeros in equality constraint Jacobian...:  2816768
Number of nonzeros in inequality constraint Jacobian.:   516442
Number of nonzeros in Lagrangian Hessian.............:   897470

Total number of variables............................:   687301
                     variables with only lower bounds:   643410
                variables with lower and upper bounds:      327
                     variables with only upper bounds:     4894
Total number of equality constraints.................:   460287
Total number of inequality constraints...............:   219364
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:   219364

iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  4.3068916e-08 1.31e-06 5.01e-04  -9.0 0.00e+00    -  0.00e+00 0.00e+00   0 
double free or corruption (out)

signal (6): Aborted
in expression starting at /net/work/barner/GGM/GGM_main_calib_toy_small.jl:107
```

odow:
Do you have a reproducible example? How are you calling Ipopt?

LukasBarner:
[...]

PR #345, opened 2022-11-28
Create instructions for loading HSL and Pardiso libraries at runtime

LukasBarner:
x-ref: https://github.com/jump-dev/Ipopt.jl/issues/342
Maybe it might make sense to run a conditional and something like this from inside the wrapper?

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,7 @@ these can be tricky, however, the following instructions should work. If they
 don't, or are not explicit enough, please open an issue.
 
 ### General Remarks
+
 During the procedures below, you may encounter errors similar to the following one:
 ```
 Exception of type: DYNAMIC_LIBRARY_FAILURE in file "Common/IpLibraryLoader.cpp" at line 67:
--- a/README.md
+++ b/README.md
@@ -182,6 +182,24 @@ To improve performance, Ipopt supports a number of linear solvers. Installing
 these can be tricky, however, the following instructions should work. If they
 don't, or are not explicit enough, please open an issue.
 
+### General Remarks
+During the procedures below, you may encounter errors similar to the following one:
+```
+Exception of type: DYNAMIC_LIBRARY_FAILURE in file "Common/IpLibraryLoader.cpp" at line 67:
+ Exception message: <path/to/somewhere.so>: undefined symbol: some_symbol
+```
+One known workaround to these errors is to `dlopen()` the library related to 
+`some_symbol` with `RTLD_GLOBAL` before usage.
+For example, if the undefined symbol is `dtrmm_`, using the following snippet 
+(prior to calling Ipopt) may help:
+```julia
+using Libdl
+Libdl.dlopen("path/to/liblapack.so.3", RTLD_GLOBAL)
+```
+If the library is on your load path, you may omit the path specification.
+Note: You may have to call `dlopen()` for more than one library (e.g. Metis or OpenMP), 
+check the respective error messages. The general procedure however remains unchanged. 
+
 ### Julia 1.7
 
 Depending on your system, you may encounter the error:


README.md
```diff
@@ -182,6 +182,24 @@ To improve performance, Ipopt supports a number of linear solvers. Installing
 these can be tricky, however, the following instructions should work. If they
 don't, or are not explicit enough, please open an issue.
 
+### General Remarks
```
odow, line 185:
```suggestion
### General Remarks

```

odow (commented):
Is this not just a problem with how we're configuring and compiling the linear solvers?

I don't think we want to do this inside Ipopt.jl, because it might have unintended consequences.

cc @amontoison

amontoison:
> Is this not just a problem with how we're configuring and compiling the linear solvers?
> 
> I don't think we want to do this inside Ipopt.jl, because it might have unintended consequences.
> 
> cc @amontoison

It's an issue with how the shared libraries of the linear solvers are compiled.
Something is not correctly linked when the shared library `libcoinhsl` is generated.
You can regenerate the shared library from the static library with this command:
```shell
# Linux
gfortran -fPIC -shared -Wl,--whole-archive libcoinhsl.a -Wl,--no-whole-archive
                                        -lblas -llapack -lmetis -lgomp
                                        -o libcoinhsl.so

# MacOS
gfortran -fPIC -shared -Wl,-all_load libcoinhsl.a -Wl,-noall_load
                                        -lblas -llapack -lmetis -lgomp
                                        -o libcoinhsl.dylib
```

LukasBarner:
Thanks for the quick feedback @amontoison 😄. I tried this and got the following error: 
```
gfortran -fPIC -shared -Wl,-all_load libcoinhsl.a -Wl,-noall_load -lblas -llapack -lmetis -lgomp -o libcoinhsl.dylib

ld: warning: option -noall_load is obsolete and being ignored
ld: library not found for -lmetis
collect2: error: ld returned 1 exit status
```
Which confuses me, because something like this works: 
```julia
julia> using Libdl

julia> dlopen("libmetis.dylib",RTLD_GLOBAL)
Ptr{Nothing} @0x00007fadce505080
```
Do you have any idea why metis is not found?

amontoison:
> Thanks for the quick feedback @amontoison smile. I tried this and got the following error:
> 
> ```
> gfortran -fPIC -shared -Wl,-all_load libcoinhsl.a -Wl,-noall_load -lblas -llapack -lmetis -lgomp -o libcoinhsl.dylib
> 
> ld: warning: option -noall_load is obsolete and being ignored
> ld: library not found for -lmetis
> collect2: error: ld returned 1 exit status
> ```
> 
> Which confuses me, because something like this works:
> 
> ```julia
> julia> using Libdl
> 
> julia> dlopen("libmetis.dylib",RTLD_GLOBAL)
> Ptr{Nothing} @0x00007fadce505080
> ```
> 
> Do you have any idea why metis is not found?

No, is `libmetis.dylib` in a folder of `DYLD_FALLBACK_LIBRARY_PATH`?
You can help the linker otherwise:
```
gfortran -fPIC -shared -Wl,-all_load libcoinhsl.a -lblas
               -llapack -L/path/to/metis -lmetis -lgomp -o libcoinhsl.dylib
```

odow:
Is a better solution to tell people to use https://github.com/JuliaSmoothOptimizers/HSL.jl?

LukasBarner:
After building from http://glaros.dtc.umn.edu/gkhome/fsroot/sw/metis/OLD, [this](https://github.com/jump-dev/Ipopt.jl/pull/345#issuecomment-1331024042) also works without any additional specifications/tweaks! (except specifying `set_optimizer_attribute()`)
Is there an easier way to get metis 4 for osx than the above (homebrew only seems to work for 5)?

amontoison:
I think that the best solution is that compile `libcoinhsl` with HSL.jl such that all dependencies are Julia artifacts.
I recompiled `METIS4` with a different soname in Yggdrasil and I directly link with it in HSL.jl.
https://github.com/JuliaSmoothOptimizers/HSL.jl/pull/150

LukasBarner:
That's nice! The dlopen() way is really not a good idea... 
I believe that https://github.com/jump-dev/Ipopt.jl/issues/341 was caused by this.
It might make sense to also use the option of ipopt to load the pardiso library at runtime (I could try testing / creating instructions for this as well)...

amontoison:
I think that your segmentation fault is another problem but all solvers will be updated with the new `coinhsl` archive so it could solve the issue.
The shared library will be libhsl.so also, so we will not need to rename it anymore.

We will also have the `c` and `z` versions of the solvers. I asked it for HSL.jl.

odow:
Closing because we're about to revamp how this all works.

PR #350, opened 2023-02-26
Add GetIpoptCurrentIterate and GetIpoptCurrentViolations

### LINKED ISSUE #349: Add support for GetIpoptCurrentViolations
Asked for on SO: https://stackoverflow.com/questions/75547527/access-the-complementarity-infeasibility-dual-infeasibility-and-the-primal-inf

x-ref https://coin-or.github.io/Ipopt/IpStdCInterface_8h.html#ae505cff0e907b9049062f665dadd8be9

This is a new feature added in v3.14.0

odow:
Closes #349 

I haven't documented the usage in the README. I think these functions are reasonably advanced, so people using them should be comfortable looking at the C API for details?

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,11 @@ See the [Ipopt documentation](https://coin-or.github.io/Ipopt/OUTPUT.html) for
 an explanation of the arguments to the callback. They are identical to the
 output contained in the logging table printed to the screen.
 
+To access the current solution and primal, dual, and complementarity violations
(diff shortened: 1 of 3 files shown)


blegat:
I guess it's difficult to have an MOI attribute for that as MOI would ask for these constraint-wise and Ipopt give all of them at the same time. So we would need to cache and then invalidate the cache when Ipopt calls a callback

blegat (approved):

PR #358, opened 2023-05-10
Test Ipopt.jl with Ipopt 3.12.14

amontoison:


odow:
I really don't understand. `Ipopt_jll v300.1400.1200+0` works fine on Windows. But not the executable...

amontoison:
I will cross-compile the old release Ipopt 3.14.4 on Windows and we will test my fork of `Ipopt_jll` in `AmplNLWriter.jl`.
It will help to find the culprit.

PR #369, opened 2023-06-07
Support the new precompiled Ipopt with LBT

amontoison:
@odow

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 
 [compat]
-Ipopt_jll = "=300.1400.400, =300.1400.1201"
+Ipopt_jll = "=300.1400.400, =300.1400.1300"
 MathOptInterface = "1.17"
 OpenBLAS32_jll = "0.3.10"
 PrecompileTools = "1"
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 
 [compat]
-Ipopt_jll = "=300.1400.400, =300.1400.403, =300.1400.1000, =300.1400.1200"
+Ipopt_jll = "=300.1400.400, =300.1400.1201"
 MathOptInterface = "1.17"
 OpenBLAS32_jll = "0.3.10"
 PrecompileTools = "1"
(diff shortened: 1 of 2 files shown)


odow:
Looks like an issue with spral:
```
In SpralSolverInterface::Factorization: Unhandled error. info.flag = -53.
Maybe one forgot to set environment variable OMP_CANCELLATION to TRUE.
```

I'd rather not need to set environment variables. have you tested spral? Perhaps we could just stick with mumps as the default?

odow:
Is this going to fix our Windows binary issue?

amontoison:
> Looks like an issue with spral:
> 
> ```
> In SpralSolverInterface::Factorization: Unhandled error. info.flag = -53.
> Maybe one forgot to set environment variable OMP_CANCELLATION to TRUE.
> ```
> 
> I'd rather not need to set environment variables. have you tested spral? Perhaps we could just stick with mumps as the default?

I thought that MUMPS was still the default linear solver:
https://github.com/coin-or/Ipopt/issues/654#issuecomment-1573698491

The plan is to keep MUMPS as the default linear solver.
I can easily do a patch if it's needed but I opened a PR upstream:
https://github.com/coin-or/Ipopt/pull/677/files

SPRAL is working if we define the following environment variables before that we start Julia:
```
export OMP_CANCELLATION=TRUE
export OMP_NESTED=TRUE
export OMP_PROC_BIND=TRUE
```
but it's not user-friendly (https://github.com/ralna/spral/issues/25#issuecomment-1579819006).

This new artifacts is compiled with LBT and without some link flags that were not relevant so It could solve the issue with the binary on Windows.
Ipopt is now compiled with the optimization level `-O3` so It should faster than before.

odow:
> I can easily do a patch if it's needed but I opened a PR upstream:

Seems like we could also set the solver ourselves inside Ipopt.jl?

amontoison:
Can  you do that be default everytime that we create a JuMP model?

odow:
You could do it here in `CreateIpoptProblem`:
https://github.com/jump-dev/Ipopt.jl/blob/f66f88840c13b382500f6b0e1ce2fd6f17d26ff9/src/C_wrapper.jl#L274-L292

amontoison:
Stefan released Ipopt 3.14.13 this morning. MUMPS will be the default now.

Project.toml
```diff
@@ -11,7 +11,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 
 [compat]
-Ipopt_jll = "=300.1400.400, =300.1400.403, =300.1400.1000, =300.1400.1200"
+Ipopt_jll = "=300.1400.400, =300.1400.1201"
```
amontoison, line 14:
```suggestion
Ipopt_jll = "=300.1400.400, =300.1400.1300"
```

Issue #374, opened 2023-06-09
Crash on Windows when optimizing with SPRAL_jll


niclasmattsson:
Not sure where the problem is so I hope it's OK if I open it here. Here's a quick test of the diet problem (from JuMP docs), first solved using Ipopt/MUMPS and then an attempt with SPRAL.

I'm not 100% sure I set everything up correctly though. I set the environment variables OMP_* as noted in the Ipopt README, started Julia and entered the diet problem with JuMP, then `]add SPRAL_jll` and `import SPRAL_jll`, then finally what you see below. Did I miss anything obvious?

```
julia> optimize!(model)

******************************************************************************
This program contains Ipopt, a library for large-scale nonlinear optimization.
 Ipopt is released as open source code under the Eclipse Public License (EPL).
         For more information visit https://github.com/coin-or/Ipopt
******************************************************************************

This is Ipopt version 3.14.13, running with linear solver MUMPS 5.6.0.

Number of nonzeros in equality constraint Jacobian...:        0
Number of nonzeros in inequality constraint Jacobian.:       63
Number of nonzeros in Lagrangian Hessian.............:        0

Total number of variables............................:        9
                     variables with only lower bounds:        9
                variables with lower and upper bounds:        0
                     variables with only upper bounds:        0
Total number of equality constraints.................:        0
Total number of inequality constraints...............:        7
        inequality constraints with only lower bounds:        4
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        3

iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  1.7819982e-01 1.77e+03 9.49e-01  -1.0 0.00e+00    -  0.00e+00 0.00e+00   0
   1  2.3371190e-01 1.76e+03 9.44e-01  -1.0 2.91e+02    -  3.56e-03 3.92e-03h  1
   2  2.7286981e-01 1.75e+03 2.87e+00  -1.0 3.14e+02    -  1.18e-02 4.60e-03h  1
   3  3.6655206e-01 1.74e+03 1.04e+00  -1.0 3.08e+02    -  1.96e-03 7.99e-03h  1
   4  5.4477465e-01 1.72e+03 7.28e+00  -1.0 2.99e+02    -  3.48e-02 1.34e-02h  1
[...]

PR #377, opened 2023-07-25
Use the last version of Ipopt_jll

amontoison:
Ipopt was recompiled with the last release of `MUMPS` and `SPRAL`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 
 [compat]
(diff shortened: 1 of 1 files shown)


odow:
Will this fix the AmplNLWriter problem?

odow (approved):

amontoison:
> Will this fix the AmplNLWriter problem?

It's unlikely.
We don't know the origin of the `AmplNLWriter` problem.

odow:
What about https://github.com/jump-dev/Ipopt.jl/issues/374

amontoison:
I was in England two weeks ago to visit the HSL team and they fixed many bugs in SPRAL.
It could solve the issue on Windows. :crossed_fingers: 
I know they did other fixes recently but we don't have a new release yet.

PR #378, opened 2023-08-07
Update the version of Ipopt_jll

amontoison:
I recompiled Ipopt with the release `2023.8.2` of SPRAL.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 
 [compat]
-Ipopt_jll = "=300.1400.400, =300.1400.1301"
+Ipopt_jll = "=300.1400.400, =300.1400.1302"
 MathOptInterface = "1.17"
 OpenBLAS32_jll = "0.3.10"
 PrecompileTools = "1"
--- a/README.md
+++ b/README.md
@@ -287,6 +287,5 @@ set_attribute(model, "linear_solver", "spral")
 Note that the following environment variables must be set before starting Julia:
 ```raw
 export OMP_CANCELLATION=TRUE
-export OMP_NESTED=TRUE
 export OMP_PROC_BIND=TRUE
 ```


odow (approved):

chriscoey:
@amontoison let's merge this? I'm not sure if it will fix my issue, but I had trouble using SPRAL even though I followed the instructions [here](https://github.com/jump-dev/Ipopt.jl#spral) and exported the three environment variables in the terminal before starting Julia:
```
julia> optimize!(model)
Setting: "spral" is not a valid setting for Option: linear_solver. Check the option documentation.

### linear_solver (String)  ###
Category: Linear Solver
Description: Linear solver used for step computations.
Valid Settings:
	ma27 (load the Harwell routine MA27 from library at runtime)
	ma57 (load the Harwell routine MA57 from library at runtime)
	ma77 (load the Harwell routine HSL_MA77 from library at runtime)
	ma86 (load the Harwell routine MA86 from library at runtime)
	ma97 (load the Harwell routine MA97 from library at runtime)
	pardiso (load the Pardiso package from pardiso-project.org from user-provided library at runtime)
	mumps (use the Mumps package)
	custom (use custom linear solver (expert use))
Default: "mumps"
ERROR: IPOPT: Couldn't set option 'linear_solver' to value 'spral'.
```

odow:
@chriscoey what version of Ipopt were you using?

chriscoey:
I was using v1.4.1 - up to date. I will try again when 1.4.2 gets tagged. All I did was paste the exports from the README (`OMP_CANCELLATION=TRUE OMP_NESTED=TRUE OMP_PROC_BIND=TRUE`, though now there's only two in the README) into the console then open julia 1.9.2 and run 
```
using Ipopt, JuMP

# An optimization problem
@variable(model, x)
@objective(model, Min, (x - 2)^2)

# Use the linear solver SPRAL
set_attribute(model, "linear_solver", "spral")
optimize!(model)
```
Can we add any tests for SPRAL?

odow:
No, I mean, what version of Ipopt the solver (Ipopt_jll, or the initial part of a log file)

odow:
We didn't add tests because of the environment variable problem. But I guess we could

chriscoey:
Oh sorry, here: version = "300.1400.400+0"

chriscoey:
Upgraded to Ipopt v1.4.2, but my version of Ipopt_jll is still the same, and I still get the error above. The environment variables are correctly set:
```
ccoey@CCoey's Macbook Pro ~ % echo $OMP_CANCELLATION
TRUE
ccoey@CCoey's Macbook Pro ~ % echo $OMP_PROC_BIND   
TRUE
```
and I see this when printing `ENV` in Julia too.

odow:
I think you need to more recent version of Ipopt_jll

chriscoey:
Hm OK, what is the recommended way to keep that up to date? I naively thought (and I think users would think the same...) that updating Ipopt.jl, or doing `]up`, would keep everything up to date, but I did that and it doesn't do anything to Ipopt_jll. So do I need to manually add Ipopt_jll and update that?

odow:
JLLs are no different than any other dependency. You probably have some other compat holding you back. If you want to force a compat bound on Ipopt_jll, you'll need to add it as an explicit dependency.

chriscoey:
Weird. Perhaps it was SCIP? I just did:
```
(@v1.9) pkg> add Ipopt_jll
   Resolving package versions...
    Updating `~/.julia/environments/v1.9/Project.toml`
⌅ [9cc047cb] + Ipopt_jll v300.1400.400+0
  No Changes to `~/.julia/environments/v1.9/Manifest.toml`

(@v1.9) pkg> up
    Updating registry at `~/.julia/registries/General.toml`
   Installed Ipopt_jll ─ v300.1400.1302+0
   Installed SPRAL_jll ─ v2023.8.2+0
  Downloaded artifact: SPRAL
  Downloaded artifact: Ipopt
    Updating `~/.julia/environments/v1.9/Project.toml`
  [9cc047cb] ↑ Ipopt_jll v300.1400.400+0 ⇒ v300.1400.1302+0
    Updating `~/.julia/environments/v1.9/Manifest.toml`
⌃ [82193955] ↓ SCIP v0.11.13 ⇒ v0.11.6
  [e33a78d0] + Hwloc_jll v2.9.2+0
  [9cc047cb] ↑ Ipopt_jll v300.1400.400+0 ⇒ v300.1400.1302+0
  [d7ed1dd3] ↑ MUMPS_seq_jll v5.4.1+0 ⇒ v500.600.100+0
⌅ [fc9abe76] ↓ SCIP_PaPILO_jll v800.0.301+0 ⇒ v0.1.0+3
⌅ [e5ac4fe4] ↓ SCIP_jll v800.0.301+0 ⇒ v0.2.1+0
  [319450e9] + SPRAL_jll v2023.8.2+0
        Info Packages marked with ⌃ and ⌅ have new versions available, but those with ⌅ are restricted by compatibility constraints from upgrading. To see why use `status --outdated -m`
```

chriscoey:
So SCIP's project.toml has ipopt_jll as a dep and no version bound, but also SCIP_jll has ipopt_jll as a dep plus a version upper bound. I think we should try to change SCIP_jll...

chriscoey:
SCIP_PaPILO_jll does the same to ipopt_jll

chriscoey:
And I can use SPRAL now that Ipopt_jll is updated 🥳

PR #389, opened 2023-10-11
Add support for MOI.BarrierIterations

### LINKED ISSUE #388: Get number of iterations after optimization
It would be great to have a way to retrieve the number of iterations that Ipopt performed during an optimization. This can be useful to understand whether a particular cost function is either slow to evaluate or slow to converge. It can also be used to have a more fine-grained comparison between optimizers.

odow:
Closes #388

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -912,7 +912,7 @@ function MOI.optimize!(model::Optimizer)
         inner.mult_x_U[i] = _dual_start(model, model.mult_x_U[i], -1)
     end
     model.barrier_iterations = 0
-    function _default_moi_callback(args...)
+    function _moi_callback(args...)
         # iter_count is args[2]
         model.barrier_iterations = max(model.barrier_iterations, args[2])
         if model.callback !== nothing
@@ -920,7 +920,7 @@ function MOI.optimize!(model::Optimizer)
         end
         return true
     end
-    SetIntermediateCallback(inner, _default_moi_callback)
+    SetIntermediateCallback(inner, _moi_callback)
     IpoptSolve(inner)
     model.solve_time = time() - start_time
     return
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -42,6 +42,7 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     qp_data::QPBlockData{Float64}
     nlp_model::Union{Nothing,MOI.Nonlinear.Model}
     callback::Union{Nothing,Function}
+    barrier_iterations::Int
 
     function Optimizer()
         return new(
@@ -63,6 +64,7 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
             QPBlockData{Float64}(),
             nothing,
             nothing,
+            0,
         )
     end
 end
@@ -105,6 +107,7 @@ function MOI.empty!(model::Optimizer)
     model.nlp_model = nothing
     model.qp_data = QPBlockData{Float64}()
     model.callback = nothing
+    model.barrier_iterations = 0
     return
 end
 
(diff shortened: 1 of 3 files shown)


src/MOI_wrapper.jl
```diff
@@ -908,9 +911,16 @@ function MOI.optimize!(model::Optimizer)
         inner.mult_x_L[i] = _dual_start(model, model.mult_x_L[i])
         inner.mult_x_U[i] = _dual_start(model, model.mult_x_U[i], -1)
     end
-    if model.callback !== nothing
-        SetIntermediateCallback(inner, model.callback)
+    model.barrier_iterations = 0
+    function _default_moi_callback(args...)
```
ojwoodford, line 915: "default" implies it can be replaced. But it can't be. It's wraps the user callback. Why not just _moi_callback()?

src/MOI_wrapper.jl
```diff
@@ -908,9 +911,16 @@ function MOI.optimize!(model::Optimizer)
...
+        # iter_count is args[2]
+        model.barrier_iterations = max(model.barrier_iterations, args[2])
```
ojwoodford, line 917: Why max() here? Is there a reason to believe Ipopt would decrease this number? What would happen if it did?

ojwoodford (commented):
Thanks for doing this. Thought I'd comment. Hope it's useful.

src/MOI_wrapper.jl
odow, line 915:
```suggestion
    function _moi_callback(args...)
```

src/MOI_wrapper.jl
```diff
@@ -908,9 +911,16 @@ function MOI.optimize!(model::Optimizer)
...
+        if model.callback !== nothing
+            return model.callback(args...)
+        end
+        return true
     end
+    SetIntermediateCallback(inner, _default_moi_callback)
```
odow, line 923:
```suggestion
    SetIntermediateCallback(inner, _moi_callback)
```

src/MOI_wrapper.jl
odow: I put max to be safe. I don't really know how Ipopt manages the intermediate callback. My guess is that it's safe to do `model.barrier_iterations = args[2]`, but I'm not sure.

src/MOI_wrapper.jl
```diff
@@ -908,9 +911,16 @@ function MOI.optimize!(model::Optimizer)
         inner.mult_x_L[i] = _dual_start(model, model.mult_x_L[i])
         inner.mult_x_U[i] = _dual_start(model, model.mult_x_U[i], -1)
     end
-    if model.callback !== nothing
-        SetIntermediateCallback(inner, model.callback)
+    model.barrier_iterations = 0
+    function _moi_callback(args...)
+        # iter_count is args[2]
+        model.barrier_iterations = max(model.barrier_iterations, args[2])
```
odow, line 917:
```suggestion
        model.barrier_iterations = args[2]
```

src/MOI_wrapper.jl
```diff
@@ -908,9 +911,16 @@ function MOI.optimize!(model::Optimizer)
         inner.mult_x_L[i] = _dual_start(model, model.mult_x_L[i])
         inner.mult_x_U[i] = _dual_start(model, model.mult_x_U[i], -1)
     end
-    if model.callback !== nothing
-        SetIntermediateCallback(inner, model.callback)
+    model.barrier_iterations = 0
+    function _moi_callback(args...)
+        # iter_count is args[2]
+        model.barrier_iterations = args[2]
```
odow, line 917: Docs say that this method is called exactly once per iteration during convergence, so I guess it is save to do this: 

https://coin-or.github.io/Ipopt/classIpopt_1_1TNLP.html#a2f962a4c43464adb7928771af84503d6

src/MOI_wrapper.jl
odow: Fixed

PR #391, opened 2023-10-25
Some performance and allocation optimizations

yuyichao:
https://github.com/jump-dev/Ipopt.jl/pull/357 appears to increase the memory allocation for a problem I have by ~4x in number and ~1.7x in size. (It doesn't affect the constructed problem but it was affecting the overall performance on a somewhat memory constraint system (relative to the problem size))

Most of the allocations seems to be caused by unnecessary allocation of the array which is fixed in the second commit. The first and the third commits are some minor and easy optimizations I noticed while going through the code. These changes got the final allocation to be ~2x in number and ~1.1x in size.

The remaining additional allocation comes from the use of quadratic function to uniformly store all the constraints and objective information. (The extra allocation has two parts, one from the creation of the empty quadratic term vector when converting to a quadratic function and the other one from the index mapping for those terms at https://github.com/jump-dev/MathOptInterface.jl/blob/36ed9d8c8561d2fdc0cb7c1fe773566cbc0c7e29/src/Utilities/functions.jl#L315 ). This is fixed in the last commit by making the functions unions and rely on the union splitting from the compiler to deal with the resulting type instability. With this last commit, I got an allocation count/size that's basically the same as before.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -240,22 +240,17 @@ function eval_sparse_gradient(
     return i
 end
 
-function append_sparse_hessian_structure!(
-    f::MOI.ScalarQuadraticFunction,
-    indices,
-)
+function append_sparse_hessian_structure!(f::MOI.ScalarQuadraticFunction, H)
     for term in f.quadratic_terms
         if _is_parameter(term.variable_1) || _is_parameter(term.variable_2)
             continue
         end
-        push!(indices, (term.variable_1.value, term.variable_2.value))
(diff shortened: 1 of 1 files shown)


odow:
Nice. I'll fix the formatting. I went for simplicity over performance with #357, and I figured it didn't matter because no one complained (until now :smile:).

odow (approved):

odow:
Still happy?

yuyichao:
Yeah looks good to me.

PR #394, opened 2023-11-16
Upgrade Ipopt_jll.jl

amontoison:


Project.toml
```diff
@@ -11,7 +11,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 
 [compat]
-Ipopt_jll = "=300.1400.400, =300.1400.1302"
+Ipopt_jll = "=300.1400.400, =300.1400.1303"
```
odow, line 14:
```suggestion
Ipopt_jll = "=300.1400.400, =300.1400.1302, =300.1400.1303"
```
Is there any reason not to go

amontoison: We did this modification with the last 5 updates of `Ipopt_jll.jl` Oscar.
One of the main reason is that we will never test again the version `300.1400.1302` with CI if the version `300.1400.1303` is allowed.

odow: > We did this modification with the last 5 updates of Ipopt_jll.jl Oscar

😆

odow:
Not sure why FreeBSD failed.

PR #395, opened 2023-11-30
Add a test for HSL solvers

amontoison:


test/MOI_wrapper.jl
```diff
@@ -7,6 +7,7 @@ module TestMOIWrapper
 
 using Ipopt
 using Test
+using HSL_jll
```
odow, line 10:
```suggestion
import HSL_jll
```

Project.toml
```diff
@@ -11,6 +11,7 @@ OpenBLAS32_jll = "656ef2d0-ae68-5445-9ca0-591084a874a2"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 
 [compat]
+HSL_jll = "2"
```
amontoison, line 14: It will not work if we have the official `HSL_jll.jl`. The version number is `2023.11.7`.

PR #406, opened 2024-02-28
Error when new and old nonlinear APIs are mixed

odow:
x-ref https://github.com/jump-dev/ParametricOptInterface.jl/issues/146


src/MOI_wrapper.jl
```diff
@@ -141,14 +141,23 @@ function MOI.supports_add_constrained_variable(
     return true
 end
 
+
```
odow, line 144:
```suggestion
```

PR #435, opened 2024-11-23
Check primal feasibility after termination failure

Robbybp:
Quick patch implementing my proposal in https://github.com/jump-dev/Ipopt.jl/issues/436

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -1141,36 +1141,6 @@ end
 
 ### MOI.TerminationStatus
 
-function _termination_status_from_restoration_failure(model::Optimizer)
-    x = model.inner.x
-    x_lbs = model.variables.lower
-    x_ubs = model.variables.upper
-    g = model.inner.g
-    g_lbs = copy(model.qp_data.g_L)
-    g_ubs = copy(model.qp_data.g_U)
-    # Assuming constraints are guaranteed to be in the order [qp_cons, nlp_cons]
-    for bound in model.nlp_data.constraint_bounds
-        push!(g_lbs, bound.lower)
-        push!(g_ubs, bound.upper)
-    end
-    # 1e-8 is the default tolerance
-    tol = get(model.options, "tol", 1e-8)
-    # 1e-6 is the default acceptable tolerance
-    atol = get(model.options, "acceptable_tol", 1e-6)
-    # TODO: Distinguish between equality and inequality constraints here?
-    bounds_feasible = all((x_lbs .- tol) .<= x .<= (x_ubs .+ tol))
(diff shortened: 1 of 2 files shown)


odow:
Do we have a trivial example that terminates with `:Restoration_Failure`?

src/MOI_wrapper.jl
```diff
@@ -1202,6 +1202,31 @@ end
...
+       all(g_L[i] - tol <= g[i] <= g_U[i] + tol for i in 1:m)
+        return MOI.FEASIBLE_POINT
+    end
+    # 1e-6 is the default acceptable tolerance
+    atol = get(model.options, "acceptable_tol", 1e-6)
+    if all(x_L[i] - atol <= x[i] <= x_U[i] + atol for i in 1:n) &&
+       all(g_L[i] - atol <= g[i] <= g_U[i] + atol for i in 1:m)
+        return MOI.NEARLY_FEASIBLE_POINT
+    end
+    return MOI.LOCALLY_INFEASIBLE
```
odow, line 1227:
```suggestion
    return MOI.INFEASIBLE_POINT
```

odow (approved):
Happy?

Robbybp:
Yes, checking this for `PrimalStatus` makes more sense than `TerminationStatus`. `LOCALLY_SOLVED` does seem a little optimistic for `RestorationFailure`, but I'd argue that `SLOW_PROGRESS` or `OTHER_ERROR` are better descriptions of what's happening than `NUMERICAL_ERROR`.

From the `LOCALLY_SOLVED` [documentation](https://jump.dev/JuMP.jl/stable/api/JuMP/#JuMP.LOCALLY_SOLVED), "could not find directions for improvement" seems to be as good a description as any for feasible restoration failure.

I'll leave this up to you. I think my preference is to return `OTHER_ERROR` for a restoration failure, but I'm not sure how big of a deal this is from a backwards compatibility point of view.

Robbybp:
> Do we have a trivial example that terminates with `:Restoration_Failure`?

This is hard. I don't have an example that does, and an algebraic example would probably be pretty sensitive to numerics or algorithm changes. If anybody has such a problem, I'd be interested to see it. We might be able to hack together some fake problem with bad derivative information in the C interface that does this, but I don't think this helps us test the MOI interface.

odow:
I'm okay with `OTHER_ERROR`. `SLOW_PROGRESS` is also good to me:
https://github.com/jump-dev/MathOptInterface.jl/blob/02dc52c05c315cb99c6392387dc0ca05d75d484b/src/attributes.jl#L2140-L2141

I don't regard it as a breaking change.

Robbybp:
LGTM

PR #447, opened 2024-11-25
Fix return type of `Bool` for ccall methods

odow:


src/C_wrapper.jl
```diff
@@ -325,15 +301,12 @@ function AddIpoptNumOption(prob::IpoptProblem, keyword::String, value::Float64)
     if !isascii(keyword)
         error("IPOPT: Non ASCII parameters not supported")
     end
-    ret = ccall(
-        (:AddIpoptNumOption, libipopt),
-        Cint,
```
odow, line 330: A note to my future self when I come looking: the C API has `bool` as a return value. We don't have `Cbool` in Julia. But using `Cint` is dangerous because it sets only the `0/1` bit, leaving all others as random data.

odow: Indeed, this has been around since the dawn of time: https://github.com/jump-dev/Ipopt.jl/blob/323fb563e3333b6c23fc479d5ec98f533f5190c2/src/Ipopt.jl#L202
Back then it was a much simpler time, when dictionaries used the Python syntax:
https://github.com/jump-dev/Ipopt.jl/blob/323fb563e3333b6c23fc479d5ec98f533f5190c2/src/Ipopt.jl#L50-L69

PR #457, opened 2025-02-12
Fix type stability of MOI nonlinear callbacks

odow:


src/MOI_wrapper.jl
```diff
@@ -891,7 +891,10 @@ function MOI.jacobian_structure(model::Optimizer)
     J = MOI.jacobian_structure(model.qp_data)
     offset = length(model.qp_data)
     if length(model.nlp_data.constraint_bounds) > 0
-        for (row, col) in MOI.jacobian_structure(model.nlp_data.evaluator)
+        J_nlp = MOI.jacobian_structure(
+            model.nlp_data.evaluator,
+        )::Vector{Tuple{Int,Int}}
```
odow, line 896:
```suggestion
        )::Vector{Tuple{Int64,Int64}}
```

PR #466, opened 2025-03-24
Add Ipopt._VectorNonlinearOracle <: MOI.AbstractVectorSet

### LINKED ISSUE #463: VectorNonlinearOracle
This is hopefully a summary of a good chat that I had with @Robbybp today.

## Motivation

We have a standard Ipopt NLP:
```
min  f(x, y)
s.t. l <= g(x, y) <= u
```
that we want to add a vector-valued nonlinear function to:
```
min  f(x, y)
s.t. l <= g(x, y) <= u
     y = F(x)
```

Vector-valued user-defined functions crop up all the time on the forum, especially in the optimal control community. Currently we tell them to follow https://jump.dev/JuMP.jl/dev/tutorials/nonlinear/tips_and_tricks/, which is pretty hacky.

We currently support this in MathOptAI: https://github.com/lanl-ansi/MathOptAI.jl/blob/53121ff5fbdae73f4750272d2545000bf5c67e97/src/predictors/GrayBox.jl#L7-L140

This works, but the downside is that when computing the Hessian-of-the-Lagrangian we compute one (dense) Hessian for each row in the output of `y`. The Hessian evaluation is a bottleneck in @Robbybp's work. 

He has a solution that hacks <s>abuses</s> the internal API of Ipopt.jl to take the `y = F(x)` constraints onto the end of the NLP evaluator, and there is a fast way in PyTorch to compute $u' F_{xx}(x)$ which is used in `eval_hessian_lagrangian`. But it'd be nice if there was first-class support in Ipopt.

## Solution

Define a new set in Ipopt.jl
```julia
struct VectorNonlinearOracle <: MOI.AbstractVectorSet
    input_dimension::Int
    output_dimension::Int
    f::Function
    jacobian::Function
    jacobian_structure::Vector{Tuple{Int,Int}}
    hessian_of_the_lagrangian::Function
    hessian_structure::Vector{Tuple{Int,Int}}
end
```

Make Ipopt support
```julia
MOI.add_constraint(::Optimizer, ::MOI.VectorOfVariables, ::MOI.VectorNonlinearOracle)
```
This would tack on the appropriate calls to
https://github.com/jump-dev/Ipopt.jl/blob/6bde96f874046a6d44dcead299b7c6c12d7bd39d/src/MOI_wrapper.jl#L919-L925

Here's the API I have in mind:
```julia
using JuMP, Ipopt
set = Ipopt.VectorNonlinearOracle(;
    input_dimension = 3,
    output_dimension = 2,
    f = (g, x) -> begin
        g[1] = x[1]^2
        g[2] = x[2]^2 + x[3]^3
        return
    end,
    jacobian = (J, x) -> begin
        J[1] = 2 * x[1]
        J[2] = 2 * x[2]
        J[3] = 3 * x[3]^2
        return
    end,
    jacobian_structure = [(1, 1), (2, 2), (2, 3)],
    hessian_of_the_lagrangian = (H, u, x) -> begin
        H[1] = 2 * u[1]
        H[2] = 2 * u[2]
        H[3] = 6 * x[3] * u[2]
        return
    end,
    hessian_of_the_lagrangian = [(1, 1), (2, 2), (3, 3)],
)
@variable(model, x[1:3])
@variable(model, y[1:2])
@constraint(model, [x; y] in set)
optimize!(model)
x_v = value.(x)
@test value.(y) == [x_v[1]^2, x_v[2]^2 + x_v[3]^3]
```

## Related

There is an open issue in MOI to add vector-valued nonlinear expressions: https://github.com/jump-dev/MathOptInterface.jl/issues/2402. The main thrust is vector-input, but vector-output is also desirable. 

If this works, I think it would solve most of the complains for vector-valued nonlinear, and allow us to punt on the implementation a little longer.

## Alternatives

We might consider
```
F(x) = 0
```
this would implicitly allow
```
g(y) = F(x)
```
but it would mean the standard "I have a user-defined function" in JuMP is more complicated.

odow:
Closes #463

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -1159,7 +1159,7 @@ function _eval_constraint(
     for i in 1:s.input_dimension
         s.x[i] = x[f.variables[i].value]
     end
-    ret = view(g, offset.+(1:s.output_dimension))
+    ret = view(g, offset .+ (1:s.output_dimension))
     s.f(ret, s.x)
     return offset + s.output_dimension
 end
@@ -1217,7 +1217,7 @@ function _eval_constraint_jacobian(
         s.x[i] = x[f.variables[i].value]
     end
     nnz = length(s.jacobian_structure)
-    s.jacobian(view(values, offset.+(1:nnz)), s.x)
+    s.jacobian(view(values, offset .+ (1:nnz)), s.x)
     return offset + nnz
 end
 
@@ -1267,8 +1267,8 @@ function _eval_hessian_lagrangian(
         s.x[i] = x[f.variables[i].value]
     end
     H_nnz, μ_nnz = length(s.hessian_lagrangian_structure), s.output_dimension
-    H_view = view(H, H_offset.+(1:H_nnz))
-    μ_view = view(μ, μ_offset.+(1:μ_nnz))
+    H_view = view(H, H_offset .+ (1:H_nnz))
+    μ_view = view(μ, μ_offset .+ (1:μ_nnz))
     s.hessian_lagrangian(H_view, s.x, μ_view)
     return H_offset + H_nnz, μ_offset + μ_nnz
 end
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -772,6 +772,7 @@ function MOI.add_constraint(
     f::F,
     s::S,
 ) where {F<:MOI.VectorOfVariables,S<:VectorNonlinearOracle}
+    model.inner = nothing
     push!(model.vector_nonlinear_oracle_constraints, (f, s))
     n = length(model.vector_nonlinear_oracle_constraints)
     return MOI.ConstraintIndex{F,S}(n)
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -16,7 +16,7 @@ function _is_parameter(term::MOI.ScalarQuadraticTerm)
 end
 
 """
-    VectorNonlinearOracle(;
+    _VectorNonlinearOracle(;
         dimension::Int,
         l::Vector{Float64},
         u::Vector{Float64},
@@ -33,6 +33,11 @@ S = \\{x \\in \\mathbb{R}^{dimension}: l \\le f(x) \\le u\\}
 ```
 where ``f`` is defined by the vectors `l` and `u`, and the callback oracles
 `f`, `jacobian`, and `hessian_lagrangian`.
+
+!!! warning
+    This set is experimental. We will decide by September 30, 2025, whether to
+    convert this into the public `Ipopt.VectorNonlinearOracle`, move it to
+    `MOI.VectorNonlinearOracle`, or remove it completely.
 
 ## f
 
@@ -87,7 +92,7 @@ do
 ```jldoctest
 julia> import Ipopt
 
-julia> set = Ipopt.VectorNonlinearOracle(;
+julia> set = Ipopt._VectorNonlinearOracle(;
            dimension = 3,
            l = [0.0, 0.0],
            u = [1.0, 0.0],
@@ -114,7 +119,7 @@ julia> set = Ipopt.VectorNonlinearOracle(;
        );
 ```
 """
-struct VectorNonlinearOracle <: MOI.AbstractVectorSet
+struct _VectorNonlinearOracle <: MOI.AbstractVectorSet
     input_dimension::Int
     output_dimension::Int
     l::Vector{Float64}
@@ -127,7 +132,7 @@ struct VectorNonlinearOracle <: MOI.AbstractVectorSet
     # Temporary storage
     x::Vector{Float64}
 
-    function VectorNonlinearOracle(;
+    function _VectorNonlinearOracle(;
         dimension::Int,
         l::Vector{Float64},
         u::Vector{Float64},
@@ -155,9 +160,9 @@ struct VectorNonlinearOracle <: MOI.AbstractVectorSet
     end
 end
 
-MOI.dimension(s::VectorNonlinearOracle) = s.input_dimension
+MOI.dimension(s::_VectorNonlinearOracle) = s.input_dimension
 
-MOI.copy(s::VectorNonlinearOracle) = s
+MOI.copy(s::_VectorNonlinearOracle) = s
 
 """
     Optimizer()
@@ -190,7 +195,7 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
     ad_backend::MOI.Nonlinear.AbstractAutomaticDifferentiation
 
     vector_nonlinear_oracle_constraints::Vector{
-        Tuple{MOI.VectorOfVariables,VectorNonlinearOracle},
+        Tuple{MOI.VectorOfVariables,_VectorNonlinearOracle},
     }
 
     function Optimizer()
@@ -216,7 +221,7 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
             nothing,
             0,
             MOI.Nonlinear.SparseReverseMode(),
-            Tuple{MOI.VectorOfVariables,VectorNonlinearOracle}[],
+            Tuple{MOI.VectorOfVariables,_VectorNonlinearOracle}[],
         )
     end
 end
@@ -397,7 +402,7 @@ function MOI.get(model::Optimizer, attr::MOI.ListOfConstraintTypesPresent)
     append!(ret, MOI.get(model.qp_data, attr))
     _add_scalar_nonlinear_constraints(ret, model.nlp_model)
     if !isempty(model.vector_nonlinear_oracle_constraints)
-        push!(ret, (MOI.VectorOfVariables, VectorNonlinearOracle))
+        push!(ret, (MOI.VectorOfVariables, _VectorNonlinearOracle))
     end
     return ret
 end
@@ -791,19 +796,19 @@ function MOI.set(
     return
 end
 
-### MOI.VectorOfVariables in VectorNonlinearOracle
+### MOI.VectorOfVariables in _VectorNonlinearOracle
 
 function MOI.supports_constraint(
     ::Optimizer,
     ::Type{MOI.VectorOfVariables},
-    ::Type{VectorNonlinearOracle},
+    ::Type{_VectorNonlinearOracle},
 )
     return true
 end
 
 function MOI.is_valid(
     model::Optimizer,
-    ci::MOI.ConstraintIndex{MOI.VectorOfVariables,VectorNonlinearOracle},
+    ci::MOI.ConstraintIndex{MOI.VectorOfVariables,_VectorNonlinearOracle},
 )
     return 1 <= ci.value <= length(model.vector_nonlinear_oracle_constraints)
 end
@@ -811,7 +816,7 @@ end
 function MOI.get(
     model::Optimizer,
     attr::MOI.ListOfConstraintIndices{F,S},
-) where {F<:MOI.VectorOfVariables,S<:VectorNonlinearOracle}
+) where {F<:MOI.VectorOfVariables,S<:_VectorNonlinearOracle}
     n = length(model.vector_nonlinear_oracle_constraints)
     return MOI.ConstraintIndex{F,S}.(1:n)
 end
@@ -819,7 +824,7 @@ end
 function MOI.get(
     model::Optimizer,
     attr::MOI.NumberOfConstraints{F,S},
-) where {F<:MOI.VectorOfVariables,S<:VectorNonlinearOracle}
+) where {F<:MOI.VectorOfVariables,S<:_VectorNonlinearOracle}
     return length(model.vector_nonlinear_oracle_constraints)
 end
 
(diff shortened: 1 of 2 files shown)


odow:
@Robbybp here you go:
```Julia
using Revise
ENV["JULIA_CONDAPKG_BACKEND"] = "Current"

using JuMP
import Ipopt
import PythonCall

function VectorNonlinearOracle(
    filename::String,
    x::Vector;
    device::String = "cpu",
)
    torch = PythonCall.pyimport("torch")
    torch_model = torch.load(filename; weights_only = false)
    torch_model = torch_model.to(device)
    J = torch.func.jacrev(torch_model)
    H = torch.func.hessian(torch_model)
    input_dimension = length(x)
    y = torch_model(torch.zeros(length(x)))
    output_dimension = PythonCall.pyconvert(Int, PythonCall.pybuiltins.len(y))
    function eval_f(ret::AbstractVector, x::AbstractVector)
        py_x = torch.tensor(x; device)
        py_value = torch_model(py_x).detach().cpu().numpy()
        value = PythonCall.pyconvert(Vector, py_value)
        for i in 1:length(value)
            ret[i] = value[i]
        end
        return
    end
    function eval_jacobian(ret::AbstractVector, x::AbstractVector)
        py_x = torch.tensor(x; device)
        py_value = J(py_x).detach().cpu().numpy()
        value = PythonCall.pyconvert(Matrix, py_value)
        for i in 1:length(value)
            ret[i] = value[i]
        end
        return
    end
    function eval_hessian_lagrangian(ret, x, u)
        py_x = torch.tensor(x; device)
        py_H = H(py_x)
        py_u = torch.tensor(u; device, requires_grad = false)
        result = torch.einsum("i,ijk->jk", py_u, py_H)
        hessian = PythonCall.pyconvert(Matrix, result.detach().cpu().numpy())
        k = 0
        for j in 1:length(x)
            for i in 1:j
                k += 1
                ret[k] = hessian[i, j]
            end
        end
        return
    end
    return Ipopt.VectorNonlinearOracle(;
        input_dimension,
        output_dimension,
        f = eval_f,
        jacobian_structure = Tuple{Int64,Int64}[
            (r, c) for c in 1:input_dimension for r in 1:output_dimension
        ],
        jacobian = eval_jacobian,
        hessian_lagrangian_structure = Tuple{Int64,Int64}[
            (i, j) for j in 1:input_dimension for i in 1:input_dimension if j >= i
        ],
        hessian_lagrangian = eval_hessian_lagrangian,
    )
end

function test_sigmoid()
    dir = mktempdir()
    filename = joinpath(dir, "model_Sigmoid.pt")
    PythonCall.pyexec(
        """
        import torch

        model = torch.nn.Sequential(
            torch.nn.Linear(3, 16),
            torch.nn.Sigmoid(),
            torch.nn.Linear(16, 2),
        )

        torch.save(model, filename)
        """,
        @__MODULE__,
        (; filename = filename),
    )
    set = VectorNonlinearOracle(filename, rand(3));
    @variable(model, x[i = 1:3] == i)
    @variable(model, y[1:2])
    @constraint(model, [x; y] in set)
    optimize!(model)
    assert_is_solved_and_feasible(model)
    torch = PythonCall.pyimport("torch")
    torch_model = torch.load(filename; weights_only = false)
    input = torch.tensor(value.(x))
    py_y = PythonCall.pyconvert(Vector, torch_model(input).detach().numpy())
    @test isapprox(value.(y), py_y; atol = 1e-4)
    return
end
```

Robbybp:
The following breaks for me with a dimension mismatch:
```julia
function test_sigmoid()
    dir = mktempdir()
    filename = joinpath(dir, "model_Sigmoid.pt")
    PythonCall.pyexec(
        """
        import torch

        model = torch.nn.Sequential(
            torch.nn.Linear(3, 16),
            torch.nn.Sigmoid(),
            torch.nn.Linear(16, 2),
        )

        torch.save(model, filename)
        """,
        @__MODULE__,
        (; filename = filename),
    )
    set = VectorNonlinearOracle(filename, rand(3));
    @variable(model, x[i = 1:3] == i)
    @constraint(model, x[1]*x[2]^1.23 <= 4)
    @variable(model, y[1:2])
    @constraint(model, [x; y] in set)
    optimize!(model)
    assert_is_solved_and_feasible(model)
    torch = PythonCall.pyimport("torch")
    torch_model = torch.load(filename; weights_only = false)
    input = torch.tensor(value.(x))
    py_y = PythonCall.pyconvert(Vector, torch_model(input).detach().numpy())
    @test isapprox(value.(y), py_y; atol = 1e-4)
    return
end
```

src/MOI_wrapper.jl
```diff
@@ -940,25 +1169,84 @@ function MOI.jacobian_structure(model::Optimizer)
...
+    μ::AbstractVector,
+    μ_offset::Int,
+    f::MOI.VectorOfVariables,
+    s::VectorNonlinearOracle,
+)
+    for i in 1:s.input_dimension
+        s.x[i] = x[f.variables[i].value]
+    end
+    H_view = view(H, H_offset:length(H))
+    μ_view = view(μ, μ_offset:length(μ))
```
Robbybp, line 1234: I think we need to restrict $\mu$ to only contain coordinates of the VectorNonlinearOracle's outputs:
```suggestion
    H_view = view(H, H_offset:(H_offset + length(s.hessian_lagrangian_structure) - 1))
    μ_view = view(μ, μ_offset:(μ_offset + s.output_dimension - 1))
```

odow:
Hmm okay. can reproduce. It's because of how I'm passing the `view` of the vectors between Julia and Python. I'll fix. 

Larger question: should we keep
```
[x; y] in NonlinearOracle(g) <=> g(x) == y
```
or change to:
```
x in NonlinearOracle(g, l, u) <=> l <= g(x) <= u
```

Robbybp:
I'm fine with either. The former is less code for the (typical?) user, the latter is less code for Ipopt.jl and allows more flexibility. For an "advanced feature" like this, I would typically go with the latter.

src/MOI_wrapper.jl
Robbybp: I guess you could handle this in the oracle's callbacks, since you know the input and output dimension there as well.

Robbybp: Regardless, with this fixed, this works for my transient-AC networks.

odow:
```julia
using Revise
ENV["JULIA_CONDAPKG_BACKEND"] = "Current"

using JuMP
import Ipopt
import PythonCall

function VectorNonlinearOracle(
    filename::String,
    x::Vector;
    device::String = "cpu",
)
    torch = PythonCall.pyimport("torch")
    torch_model = torch.load(filename; weights_only = false)
    torch_model = torch_model.to(device)
    J = torch.func.jacrev(torch_model)
    H = torch.func.hessian(torch_model)
    input_dimension = length(x)
    y = torch_model(torch.zeros(length(x)))
    output_dimension = PythonCall.pyconvert(Int, PythonCall.pybuiltins.len(y))
    # We model the function as:
    #     0 <= f(x) - y <= 0
    function eval_f(ret::AbstractVector, x::AbstractVector)
        py_x = torch.tensor(x[1:input_dimension]; device)
        py_value = torch_model(py_x).detach().cpu().numpy()
        value = PythonCall.pyconvert(Vector, py_value)
        for i in 1:output_dimension
            ret[i] = value[i] - x[input_dimension+i]
        end
        return
    end
    # Note the order of the for-loops, first over the output_dimension, and then
    # across the input_dimension. This makes the Jacobian structure of ∇f(x) be
    # column-major and dense with respect to x.
    jacobian_structure = Tuple{Int64,Int64}[
        (r, c) for c in 1:input_dimension for r in 1:output_dimension
    ]
    # We also need to add non-zero terms for the `-I` component of the Jacobian.
    for i in 1:output_dimension
        push!(jacobian_structure, (i, input_dimension + i))
    end
    function eval_jacobian(ret::AbstractVector, x::AbstractVector)
        py_x = torch.tensor(x[1:input_dimension]; device)
        py_value = J(py_x).detach().cpu().numpy()
        value = PythonCall.pyconvert(Matrix, py_value)
        for i in 1:length(value)
            ret[i] = value[i]             # ∇f(x)
        end
        for i in 1:output_dimension
            ret[length(value)+i] = -1.0   # -I
        end
        return
    end
    # We need to compute only ∇²f(x) because the -y part does not appear in
    # the Hessian.
    #
    # Note the order of the for-loops, first over the rows, and then across the
    # columns, with j >= i ensuring that this is the upper triangle portion of
    # the Hessian-of-the-Lagrangian.
    hessian_lagrangian_structure = Tuple{Int64,Int64}[
        (i, j) for j in 1:input_dimension for i in 1:input_dimension if j >= i
    ]
    function eval_hessian_lagrangian(ret, x, u)
        py_x = torch.tensor(x[1:input_dimension]; device)
        py_H = H(py_x)
        py_u = torch.tensor(u; device, requires_grad = false)
        result = torch.einsum("i,ijk->jk", py_u, py_H)
        hessian = PythonCall.pyconvert(Matrix, result.detach().cpu().numpy())
        k = 0
        for j in 1:input_dimension
            for i in 1:j
                k += 1
                ret[k] = hessian[i, j]
            end
        end
        return
    end
    return Ipopt.VectorNonlinearOracle(;
        dimension = input_dimension + output_dimension,
        l = zeros(output_dimension),
        u = zeros(output_dimension),
        f = eval_f,
        jacobian_structure,
        jacobian = eval_jacobian,
        hessian_lagrangian_structure,
        hessian_lagrangian = eval_hessian_lagrangian,
    )
end

function test_sigmoid()
    dir = mktempdir()
    filename = joinpath(dir, "model_Sigmoid.pt")
    PythonCall.pyexec(
        """
        import torch

        model = torch.nn.Sequential(
            torch.nn.Linear(3, 16),
            torch.nn.Sigmoid(),
            torch.nn.Linear(16, 2),
        )

        torch.save(model, filename)
        """,
        @__MODULE__,
        (; filename = filename),
    )
    set = VectorNonlinearOracle(filename, rand(3));
    @variable(model, x[i = 1:3] == i)
    @variable(model, y[1:2])
    @constraint(model, [x; y] in set)
    optimize!(model)
    assert_is_solved_and_feasible(model)
    torch = PythonCall.pyimport("torch")
    torch_model = torch.load(filename; weights_only = false)
    input = torch.tensor(value.(x))
    py_y = PythonCall.pyconvert(Vector, torch_model(input).detach().numpy())
    @test isapprox(value.(y), py_y; atol = 1e-4)
    return
end

function test_sigmoid_2()
    dir = mktempdir()
    filename = joinpath(dir, "model_Sigmoid.pt")
    PythonCall.pyexec(
        """
        import torch

        model = torch.nn.Sequential(
            torch.nn.Linear(3, 16),
            torch.nn.Sigmoid(),
            torch.nn.Linear(16, 2),
        )

        torch.save(model, filename)
        """,
        @__MODULE__,
        (; filename = filename),
    )
    set = VectorNonlinearOracle(filename, rand(3));
    @variable(model, x[i = 1:3] == i)
    @constraint(model, x[1]*x[2]^1.23 <= 4)
    @variable(model, y[1:2])
    @constraint(model, [x; y] in set)
    optimize!(model)
    assert_is_solved_and_feasible(model)
    torch = PythonCall.pyimport("torch")
    torch_model = torch.load(filename; weights_only = false)
    input = torch.tensor(value.(x))
    py_y = PythonCall.pyconvert(Vector, torch_model(input).detach().numpy())
    @test isapprox(value.(y), py_y; atol = 1e-4)
    return
end

test_sigmoid()
test_sigmoid_2()
```

odow:
I'm pretty happy with this, and I think there's a strong argument to add this to MOI directly. It's effectively an escape hatch to let you define part of a nonlinear model in JuMP with our AD etc, and then break out into a custom constraint block where you control the AD and the sparsity.

odow:
I guess we want the Hessian to be optional

src/MOI_wrapper.jl
```diff
@@ -15,6 +15,150 @@ function _is_parameter(term::MOI.ScalarQuadraticTerm)
     return _is_parameter(term.variable_1) || _is_parameter(term.variable_2)
 end
 
+"""
+    VectorNonlinearOracle(;
```
odow, line 19: Add an underscore. And a thing saying to post on the forum or open an issue in Ipopt.jl to tell us that you're using this.

odow: And include a deadline to decide whether to move to MOI or make public from Ipopt.jl. September 30.

odow: Open a GH issue with a description of how to use from JuMP so we can link people to it.

src/MOI_wrapper.jl
```diff
@@ -940,25 +1218,83 @@ function MOI.jacobian_structure(model::Optimizer)
...
+    μ_offset::Int,
+    f::MOI.VectorOfVariables,
+    s::VectorNonlinearOracle,
+)
+    for i in 1:s.input_dimension
+        s.x[i] = x[f.variables[i].value]
+    end
+    H_nnz, μ_nnz = length(s.hessian_lagrangian_structure), s.output_dimension
+    H_view = view(H, H_offset .+ (1:H_nnz))
+    μ_view = view(μ, μ_offset .+ (1:μ_nnz))
```
Robbybp, line 1283: Nitpick: `μ_nnz` is a little confusing here.
```suggestion
    H_nnz = length(s.hessian_lagrangian_structure)
    H_view = view(H, H_offset .+ (1:H_nnz))
    μ_view = view(μ, μ_offset .+ (1:s.output_dimension))
```

odow:
This seems sufficient to make https://github.com/lanl-ansi/MathOptAI.jl/pull/171 work nicely, so I'm going to merge and tag.

PR #472, opened 2025-04-14
Create a package extension for MathOptInterface.jl

amontoison:
cc @odow


Project.toml
```diff
@@ -6,23 +6,28 @@ version = "1.8.0"
...
+MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 PrecompileTools = "aea7be01-6a6a-4083-8856-8a6e6704d82a"
 
+[extensions]
+IpoptMOI = ["MathOptInterface", "PrecompileTools"]
+
 [compat]
-HSL_jll = "3, 4, 2023, 2024"
-Ipopt_jll = "=300.1400.400, =300.1400.1400, =300.1400.1600,=300.1400.1700"
+HSL_jll = "3, 4, 2023, 2024, 2025"
```
odow, line 19: Let's keep this to a different PR

odow:
Perhaps just to be very clear: I'm not against merging something like this, but it needs a lot of careful thought and testing before we do.

src/Ipopt.jl
```diff
@@ -22,7 +21,11 @@ function __init__()
 end
 
 include("C_wrapper.jl")
-include("MOI_wrapper.jl")
+
+# Functions needed by the MOI extension
+function Optimizer end
+function _VectorNonlinearOracle end
+function CallbackFunction end
```
odow, line 28: So I have issues with these: they are not functions, they are structs. Someone might have written `::Ipopt.Optimizer` or `::Ipopt.CallbackFunction` in user-code and this would break.

If we're going to do this, I want zero user-breakage.

amontoison: I started to investigate if it fixes the failing tests but it is not the case.
It is not the good approach.

odow:
Closing in favor of #478

PR #475, opened 2025-04-14
Add a function IpoptVersion

### LINKED ISSUE #473: Update MOI.get(::Optimizer, ::MOI.SolverVersion)
We should use the version of `Ipopt_jll.jl` where all numbers are divided by 100.
-> https://github.com/jump-dev/Ipopt.jl/blob/master/src/MOI_wrapper.jl#L236

amontoison:
close #473 

Update: Julia 1.9 seems required for `pkgversion`.

### DIFF
--- a/src/C_wrapper.jl
+++ b/src/C_wrapper.jl
@@ -479,3 +479,12 @@ function GetIpoptCurrentViolations(
     end
     return
 end
+
+function IpoptVersion()
+    ipopt_jll_version = pkgversion(Ipopt_jll)
+    major = ipopt_jll_version.major ÷ 100
+    minor = ipopt_jll_version.minor ÷ 100
+    patch = ipopt_jll_version.patch ÷ 100
+    ipopt_version = "$major.$minor.$patch"
+    return ipopt_version
+end
--- a/src/Ipopt.jl
+++ b/src/Ipopt.jl
@@ -33,7 +33,8 @@ export IpoptProblem,
     OpenIpoptOutputFile,
     SetIpoptProblemScaling,
     SetIntermediateCallback,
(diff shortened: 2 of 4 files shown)


src/C_wrapper.jl
```diff
@@ -479,3 +479,12 @@ function GetIpoptCurrentViolations(
     end
     return
 end
+
+function IpoptVersion()
+    ipopt_jll_version = pkgversion(Ipopt_jll)
```
odow, line 484: This requires Julia v1.9, and it also assumes a particular naming scheme of Ipopt_jll that is not guaranteed. It also does not account for `]dev`ed versions of Ipopt_jll. I just don't think we need this function.

amontoison: You can't dev an old version of `Ipopt_jll.jl` because of the compat entries.
All versions supported this particular naming scheme of Ipopt_jll.
I agree that Julia 1.9 is an issue.

odow: What is the benefit of having this? Why is it useful?

amontoison: It is just that the number is wrong if someone use it in MOI.
I don't know how useful it is for you but if we implement it, it should return something correct.
```julia
MOI.get(::Optimizer, ::MOI.SolverVersion)
```

odow: Ah I didn't look below to see

https://github.com/jump-dev/Ipopt.jl/blob/801b6a0a055f79bbf5d785f80f9ba9cdb2333ec9/src/MOI_wrapper.jl#L236

amontoison:
It's fine for me.
I close the PR.

PR #476, opened 2025-04-14
Fix MOI.SolverVersion

odow:
Closes #475


amontoison (approved):

odow:
I chose to leave the full Ipopt_jll version string. MOI doesn't make any claims about what `SolverVersion` should return. It seemed simpler than assuming some structure in the version

https://github.com/jump-dev/MathOptInterface.jl/blob/c611a46401769ce0941eff9562787b05f26f4948/src/attributes.jl#L835-L860

PR #498, opened 2025-08-08
Support Ipopt 3.14.19

amontoison:
I also updated the version of MUMPS (5.8.1).


amontoison:
@odow I think we should start to remove the support of old versions of Ipopt.
We kept 3.14.4 because of Julia 1.6 (it is the only JLL not compiled with LBT) but I think we can remove it now.

amontoison:
Another solution is to use this action: https://github.com/julia-actions/julia-downgrade-compat

PR #499, opened 2025-08-09
Define persistent @cfunction callback

amontoison:
We don't need to do `@cfunction` each time that we create `ipopt_problem`.
The callbacks are always the same.

Update: it seems that these the extension doesn't like the C function pointers outside the Julia functions. :thinking:


amontoison:
I am scared to open a can of worms for nothing. Let's keep things unchanged...

PR #500, opened 2025-08-25
Record time spent in VectorNonlinearOracle callbacks

Robbybp:


odow:
I need to think about this. I don't know that adding a timer to the set is the right option.

odow:
Closing in favor of #502

PR #508, opened 2025-10-23
Change function type from affine to quadratic

amontoison:
`objective_function_type` will be consistent with the default `objective`.


odow:
Let me try something.

PR #534, opened 2026-04-02
Add a CI job to test JuliaC

odow:
x-ref #532


odow:
Without trimming it's ~385 MB. With trimming in #535 it's ~135 MB, so there's a substantive difference, but it still pretty hefty. The main change is the `app_test_exe` binary goes from 190 MB without trimming to 2 MB with trimming. The other change is that `lib/julia` doesn't include libLLVM (60 MB).

PR #546, opened 2026-08-05
Add a test with bfpc's instance from Discourse

odow:
cc @bfpc


odow:
I'm not sure there's anything we can easily do here. This just looks like a bug in the linear algebra somewhere.

cc @amontoison

amontoison:
@odow Are we sure that the problem doesn't have a local minimum?
It seems that the 3 failing CI jobs converge to the same point.

odow:
Problem is an LP

amontoison:
It is quite hard to know what is the issue.
It may be a linear algebra bug in IPOPT or an unstable / wrong descent direction computed by the sparse linear solver.
I will suggest to use `"spral"` instead of `"mumps"` in the CI for this test to quickly see if we have the same issue.

odow:
I guess this means that it’s in MUMPS

amontoison:
Can you try again MUMPS with the compat entry `=500.900.000` ?
Maybe it is a bug in the new release but I am surprised that only the new test is failing.

Project.toml
```diff
@@ -19,6 +20,7 @@ IpoptMathOptInterfaceExt = "MathOptInterface"
 HSL_jll = "3, 4, 2023, 2024, 2025, 2026"
 Ipopt_jll = "=300.1400.1900, =300.1400.1901, =300.1400.1902"
 LinearAlgebra = "1"
+MUMPS_seq_jll = "=500.900.000"
```
odow, line 23: Looks like this isn't the culprit :cry:

bfpc:
Comparing [50be93a](https://github.com/jump-dev/Ipopt.jl/pull/546/commits/50be93ac8f1ab429075490f2faf6d2627ce0a5b9) and [b5c6dfe](https://github.com/jump-dev/Ipopt.jl/pull/546/commits/b5c6dfe98beeed190fd6144907e1cc4a4a256e67), the arch/julia combination pairs that fail are different.  Also the last one [8298885](https://github.com/jump-dev/Ipopt.jl/pull/546/commits/82988859f06d77fdc8aa04765a7b9928b00c8a41) and previous seem to fail in different version combinations.

odow:
Based on @cvanaret's comment here: https://discourse.julialang.org/t/ipopt-has-trouble-with-vector-affexpr-in-moi-zeros/138395/11?u=odow

Am I right to understand this is a bug in Ipopt/a poorly conditioned instance?

amontoison:
Yes, and it is failing with MUMPS because this sparse linear solver is better to approximate the inertia but Ipopt doesn't know how to handle more negative eigenvalues than positive ones in the KKT system.

cc @cvanaret

odow:
I'm not sure there's anything to do here from an Ipopt.jl perspective then.

odow:
Closing because I don't think there's anything to do here. @cvanaret is this something simple that could/should be fixed in Ipopt or is it relatively niche and a heavy lift?

cvanaret:
I had implemented the regularization loop described in the IPOPT implementation paper, which I guess got improved over time. I just had to test for singularity in the loop, and crank the dual regularization parameter. Should take a maintainer 5 minutes.
The Uno barrier method can now solve the instance with MA57, MA27, MA86, MUMPS and SSIDS as the linear solver.

odow:
@cvanaret can you open an issue in coin-or/Ipopt with the details?

cvanaret:
I've just tested locally with IPOPT 3.14.20, bfpc is successfully solved with (unscaled) objective 119250 with MA57, MA27, MA86 (v2026.8.4), and MUMPS (v5.9.0). I get 136285.7 as scaled objective.
Not sure what's going on here, everything looks fine...

PR #548, opened 2026-08-13
Use MOI.Nonlinear.QPBlockData

blegat:
A net line deletion of 1200 lines. Needs

- [ ] https://github.com/jump-dev/MathOptInterface.jl/pull/3048


ext/IpoptMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -1376,8 +1390,13 @@ function _setup_model(model::Optimizer)
             MOI.Nonlinear.Evaluator(model.nlp_model, model.ad_backend, vars),
         )
     end
+    model.qp_nnzj = length(MOI.jacobian_structure(model.qp_data))
+    model.qp_nnzh = length(MOI.hessian_lagrangian_structure(model.qp_data))
     has_quadratic_constraints =
-        any(isequal(_kFunctionTypeScalarQuadratic), model.qp_data.function_type)
+        any(
+            isequal(MOI.Nonlinear._kFunctionTypeScalarQuadratic),
+            model.qp_data.function_type,
```
odow: Is this part of the public API of `qp_data`?

ext/IpoptMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -193,6 +202,11 @@ function MOI.add_constrained_variable(
     push!(model.list_of_variable_indices, p)
     model.parameters[p] =
         MOI.Nonlinear.add_parameter(model.nlp_model, set.value)
+    # `QPBlockData` treats a variable as a parameter if and only if its index
+    # is a key of `parameters`, so the parameter must be registered before
+    # any structure query. The value is re-synced in `copy_parameters` before
+    # every solve.
+    model.qp_data.parameters[p.value] = set.value
```
odow: Can we improve this API somehow?

ext/IpoptMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -3,7 +3,8 @@
 # Use of this source code is governed by an MIT-style license that can be found
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
-include("utils.jl")
+const QPBlockData = MOI.Nonlinear.QPBlockData
```
odow: Strongly prefer that we use explicit prefixes instead of these extra constants. It's only needed in a few places, and using the explicit prefix makes the code much easier to read.

.github/workflows/ci.yml
```diff
@@ -41,6 +41,13 @@ jobs:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
       - uses: julia-actions/cache@v3
+      - name: MOI
+        shell: julia --project=@. {0}
+        run: |
+          using Pkg
+          Pkg.add([
+              PackageSpec(name="MathOptInterface", rev="bl/qp_block_data"),
+          ])
```
odow, line 50:
```
[sources]
MathOptInterface = {url = "https://github.com/jump-dev/MathOptInterface.jl", rev = "bl/qp_block_data"}
```

blegat: That doesn't work on Julia v1.10

ext/IpoptMathOptInterfaceExt/MOI_wrapper.jl
blegat: I think QPBlockData should be internal and we should only expose ModelWithQuad, see https://github.com/jump-dev/Ipopt.jl/pull/548

ext/IpoptMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -595,22 +520,16 @@ function MOI.get(
     model::Optimizer,
     attr::MOI.NumberOfConstraints{F,S},
 ) where {F<:MOI.ScalarNonlinearFunction,S<:_SETS}
-    if model.nlp_model === nothing
-        return 0
-    end
-    return count(v.set isa S for v in values(model.nlp_model.constraints))
+    return count(v.set isa S for v in values(model.model.inner.constraints))
```
odow: Is this public API?

ext/IpoptMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -1071,14 +987,72 @@ function MOI.set(
...
+    return MOI.initialize(d.nlp, features)
+end
+
+MOI.eval_objective(d::_OracleNLPEvaluator, x) = MOI.eval_objective(d.nlp, x)
+
+function MOI.eval_objective_gradient(d::_OracleNLPEvaluator, grad, x)
+    return MOI.eval_objective_gradient(d.nlp, grad, x)
+end
+
+function MOI.Nonlinear._constraint_bounds(d::_OracleNLPEvaluator)
```
odow: This is a private method

ext/IpoptMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -1445,13 +1421,17 @@ function MOI.optimize!(model::Optimizer)
     for i in 1:length(model.variable_primal_start)
         inner.x[i] = something(
             model.variable_primal_start[i],
-            clamp(0.0, model.variables.lower[i], model.variables.upper[i]),
+            clamp(
+                0.0,
+                model.model.variables.lower[i],
```
odow: Is this public?

ext/IpoptMathOptInterfaceExt/MOI_wrapper.jl
```diff
@@ -1623,9 +1597,11 @@ function MOI.get(
 )
     MOI.check_result_index_bounds(model, attr)
     MOI.throw_if_not_valid(model, vi)
-    if _is_parameter(vi)
-        p = model.parameters[vi]
-        return model.nlp_model[p]
+    if MOI.Nonlinear._is_parameter(vi)
```
odow, line 852: Private method

odow:
So the net deletion is more or less just the number of lines in the utils.jl file. What is EvaluatorWithQuad actually buying us?

ext/IpoptMathOptInterfaceExt/legacy.jl
```diff
@@ -0,0 +1,112 @@
...
+        error(
+            "The legacy `MOI.NLPBlock` interface cannot be combined " *
+            "with the selected automatic-differentiation backend.",
+        )
+    end
+    if model.nlp_data.has_objective
+        # The legacy NLPBlock path only supports the built-in layered model,
+        # not backend-specific models such as ExaModels.
+        @assert model.model isa MOI.Nonlinear.ModelWithQuad
+        model.model.objective_sink = MOI.Nonlinear._INNER
```
blegat, line 66: The fact that we still support this legacy is making it difficult to come up with a right design. We should just drop support for it in Ipopt.jl or accept something a bit hacky like this.

PR #549, opened 2026-08-15
Use MOI.Nonlinear.ModelWithQuad instead of QPBlockData

blegat:
Need

- [x] https://github.com/jump-dev/MathOptInterface.jl/pull/3053

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
         run: |
           using Pkg
           Pkg.add([
-              PackageSpec(name="MathOptInterface", rev="bl/qp_block_data"),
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         run: |
           using Pkg
           Pkg.add([
-              PackageSpec(name="MathOptInterface", rev="bl/qp_block_data"),
+              PackageSpec(name="MathOptInterface", rev="bl/model_with_quad"),
```
odow, line 49: Use `[sources]` in Project.toml instead

blegat: Doesn't work on Julia v1.10

